PR #56, opened 2018-01-09
add bank hapoalim scrapper

sebikaplun:
Bank hapoalim first version

### DIFF
--- a/src/definitions.js
+++ b/src/definitions.js
@@ -17,4 +17,8 @@ export const SCRAPERS = {
     name: 'Amex',
     loginFields: ['id', 'card6Digits', PASSWORD_FIELD],
   },
+  hapoalim: {
+    name: 'Bank Hapoalim',
+    loginFields: ['id', PASSWORD_FIELD],
+  },
 };
--- a/src/definitions.js
+++ b/src/definitions.js
@@ -19,6 +19,6 @@ export const SCRAPERS = {
   },
   hapoalim: {
     name: 'Bank Hapoalim',
-    loginFields: ['id', PASSWORD_FIELD],
+    loginFields: ['userCode', PASSWORD_FIELD],
   },
 };
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -71,7 +71,7 @@ function getPossibleLoginResults() {
 
 function createLoginFields(credentials) {
   return [
-    { id: 'userID', value: credentials.id },
+    { id: 'userID', value: credentials.userCode },
     { id: 'userPassword', value: credentials.password },
   ];
 }
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -65,7 +65,7 @@ function getPossibleLoginResults() {
   const urls = {};
   urls[LOGIN_RESULT.SUCCESS] = `${BASE_URL}/portalserver/HomePage`;
   urls[LOGIN_RESULT.INVALID_PASSWORD] = `${BASE_URL}/AUTHENTICATE/LOGON?flow=AUTHENTICATE&state=LOGON&errorcode=1.6&callme=false`;
-  urls[LOGIN_RESULT.CHANGE_PASSWORD] = `${BASE_URL}/portalserver/HomePage`;
+  urls[LOGIN_RESULT.CHANGE_PASSWORD] = `${BASE_URL}/MCP/START?flow=MCP&state=START&expiredDate=null`;
   return urls;
 }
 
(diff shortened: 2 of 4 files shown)


src/definitions.js
```diff
@@ -17,4 +17,8 @@ export const SCRAPERS = {
     name: 'Amex',
     loginFields: ['id', 'card6Digits', PASSWORD_FIELD],
   },
+  hapoalim: {
+    name: 'Bank Hapoalim',
+    loginFields: ['id', PASSWORD_FIELD],
```
jonathanrauch: @sebikaplun the website calls the field `קוד משתמש` so I think `id` should be `username`. `id` makes me think Teudat Zehut.

src/scrapers/hapoalim.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  return accountData;
+}
+
+function getPossibleLoginResults() {
+  const urls = {};
+  urls[LOGIN_RESULT.SUCCESS] = `${BASE_URL}/portalserver/HomePage`;
+  urls[LOGIN_RESULT.INVALID_PASSWORD] = `${BASE_URL}/AUTHENTICATE/LOGON?flow=AUTHENTICATE&state=LOGON&errorcode=1.6&callme=false`;
+  urls[LOGIN_RESULT.CHANGE_PASSWORD] = `${BASE_URL}/portalserver/HomePage`;
```
eshaham: @sebikaplun this url doesn't look right

asfktz:
Hi @eshaham and @sebikaplun!

First, I want to thank you both, you @eshaham for writing this beautiful scaper,
and you @sebikaplun for adding support for "Bank Hapoalim".
Both of you saved me so many hours of work, I appreciate it.

I wanted to suggest a few things:

### Unique identifier

Currently, the `identifier` property mapped to the transaction's `referenceNumber`

It seems like `referenceNumber ` is the last 4 digits of the credit card, but not always, it can also be some other, larger number.

Meanings that if it will be used as the `identifier` key, we're going to end up with many duplicates ids.

I tried to find some other key to use as the transition's unique `id,`  but none seems fit.

So I would like to suggest instead to generate the ID based on a mix of properties:

* `eventDate` - Date, formatted as `YYYYMMDD`

* `referenceNumber` - The 4 last digits of the credit card 

* `referenceCatenatedNumber` - Another ID of some sort, it's related to the business that initiated the transaction. 

* `eventAmount` - The Amount of the transaction

* `currentBalance` - account balance after the transaction

Not sure if it wise to relay of `currentBalance`, but if so, it can greatly add to the id's uniqueness.


### Status property

As you can see in here:
screen shot 2018-01-25 at 9 30 46 pm

Transactions above the fold are still `"pending"`
It will be useful to be able to differentiate between `pending` and `posted` transactions


```js
status: txn.serialNumber === 0 ? 'pending' : 'posted'
````

What do you think?

eshaham:
@asfktz the purpose of the `identifier` field is to represent the unique identifier / operation number the bank provides. you would usually find these in bank accounts, but not in credit card statements.
if not supplied by the bank, i don't think we should come up with a unique number of our own, as that's not really needed IMO.

as for the status property,  sounds like a good idea, though we will need to add it to the rest of the scrapers. let's do it as a part of a different PR, after @sebikaplun's PR is merged. makes sense?

asfktz:
Hi, @eshaham!

> The purpose of the identifier field is to represent the unique identifier / operation number the bank provides. you would usually find these in bank accounts, but not in credit card statements.
if not supplied by the bank, i don't think we should come up with a unique number of our own, as that's not really needed IMO.

Yeah, I see your point. The advantage of doing so is that is abstract the fact that one company provides an ID and the other don't.

It's nice to have a unique ID because it makes it easier to detect duplicate records.
But such ID can also be added by the user too.


> as for the status property, sounds like a good idea, though we will need to add it to the rest of the scrapers. let's do it as a part of a different PR, after @sebikaplun's PR is merged. makes sense?
 
Good Idea :)
I'll open a separate issue for it

eshaham:
> Yeah, I see your point. The advantage of doing so is that is abstracts away the fact that one company provides an ID and the other don't.
> 
> It's nice to have a unique ID because it makes it easier to detect duplicate records.
> But such ID can also be added by the user too.

@asfktz yes, I guess we can think about creating unique identifiers, but to have people depend on them being unique we will need to come up with a bullet proof algorithm that would work for every scraper. feel free to create an issue and start rolling that PR 😄

eshaham:
@sebikaplun notice the changes in #58, you will need to make similar changes in your scraper.

asfktz:
Hi @sebikaplun!
You did an excellent job with this scraper I and I can't wait for this PR to be merge.
I guess you're a bit short on time lately, would you like me to continue from where you left off?

src/scrapers/hapoalim.js
sebikaplun: @eshaham 5bbeabd

sebikaplun:
@eshaham updated readme 1ce52ce

sebikaplun:
@eshaham i think im ready to merge

Issue #67, opened 2018-02-08
Visa Cal scraper returning error


ifeins:
I'm using the Israeli ynab updater with the latest version of bank scrapers (0.5.3) and I'm getting the following error when trying to use the Visa Cal scraper:

```
➜  israeli-ynab-updater git:(master) ✗ npm start

> israeli-ynab-updater@0.0.1 start /Users/ifeins/Development/web/israeli-ynab-updater
> babel-node src

? Which bank would you like to scrape? Visa Cal
? Combine installment transactions? Yes
visaCal: START_SCRAPING
visaCal: LOGGING_IN
visaCal: LOGIN_SUCCESS
visaCal: END_SCRAPING
success: false
error type: generic
error: unknown transaction type 7
```

ifeins:
Found the issue, will open a PR soon.

PR #69, opened 2018-02-09
Fix visa cal scraper

ifeins:
Added support for withdrawal transaction types which without them the scraper would fail.

Fixes this issue: https://github.com/eshaham/israeli-bank-scrapers/issues/67


eshaham:
@ifeins I think you need to have the `fixes` keyword right before the issue number to actually [close the issue](https://help.github.com/articles/closing-issues-using-keywords/).
fixes #67

PR #73, opened 2018-02-12
Refactor sortTransactionsByDate

asfktz:
This PR refactor sortTransactionsByDate to use lodash.orderBy for readability

### DIFF
--- a/src/helpers/transactions.js
+++ b/src/helpers/transactions.js
@@ -1,5 +1,5 @@
 import moment from 'moment';
-import orderBy from 'lodash/orderBy';
+import _ from 'lodash';
 
 import { NORMAL_TXN_TYPE, INSTALLMENTS_TXN_TYPE } from '../constants';
 
@@ -32,7 +32,7 @@ export function fixInstallments(txns) {
 }
 
 export function sortTransactionsByDate(txns) {
-  return orderBy(txns, ['date']);
+  return _.sortBy(txns, ['date']);
 }
 
 export function filterOldTransactions(txns, startMoment, combineInstallments) {
--- a/src/helpers/transactions.js
+++ b/src/helpers/transactions.js
@@ -1,5 +1,5 @@
 import moment from 'moment';
-import orderBy from 'lodash/orderBy';
+import _ from 'lodash';
 
 import { NORMAL_TXN_TYPE, INSTALLMENTS_TXN_TYPE } from '../constants';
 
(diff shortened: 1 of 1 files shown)


src/helpers/transactions.js
```diff
@@ -30,15 +32,7 @@ export function fixInstallments(txns) {
...
-  const cloned = Array.from(txns);
-  cloned.sort((txn1, txn2) => {
-    if (txn1.date.getTime() === txn2.date.getTime()) {
-      return 0;
-    }
-    return txn1.date < txn2.date ? -1 : 1;
-  });
-
-  return cloned;
+  return orderBy(txns, 'date', 'asc');
```
eshaham: @asfktz nice touch 👍 
I think we should be using `_.orderBy(...)`, as it is simpler.
Also, both functions [seem to expect](https://lodash.com/docs#orderBy) an array for the `iteratees` param, ie, `(txns, ['date'])`.

asfktz: @eshaham thanks (:
I've changed `orderBy ` to `_.orderBy`

The iteratees can also expect a string, this is its signature:
`[iteratees=[_.identity]] (Array[]|Function[]|Object[]|string[])`

eshaham: @asfktz I'm reading my comment above and it doesn't make sense 😆 
I meant to say, use `_.sortBy(...)`, as it is simpler - no need to add `asc` param.

As for iteratees, it shows `string[]`.
Even if it may be working for you without the array, I'm not insisting just for fun.
I don't want it to break when we upgrade to a future version of lodash.

asfktz:
Hi @eshaham 😀
I made the changes

`orderBy(txns, ['date'])` instead of `_.orderBy(txns, 'date', 'asc')`

I personally prefer `orderBy` but can change it back to `_.orderBy` if you like.

eshaham:
@asfktz really don't want to sound picky, I just really prefer consistency - other place which use `lodash` in this repo use the underscore syntax. I'm open to make changes to convention, but prefer to make sure we're consistent.
Also, you missed my comment about preferring the `sortBy()` function over the `orderBy()` function.

asfktz:
@eshaham oh, I overlooked it, sorry.
And don't worry, is absolutely ok (:

Changed `orderBy(txn, ['date'], 'asc')` to `_.sortBy(txns, ['date'])`

PR #81, opened 2018-02-23
Implement multiple login results 

chilikla:
Recently, I noticed Bank Hapoalim website home page and subfolders have been different for time to time. It used to be https://login.bankhapoalim.co.il/portalserver/HomePage, but mostly recently https://login.bankhapoalim.co.il//ng-portals-bt/rb/he/homepage.
Besides that, the api urls has been changed as well (from `${BASE_URL}/portalserver`  to `${BASE_URL}/ssb`. I don't know the nature of this change, maybe it's just an A/B testing, but it failed the scarper a lot lately.

So I updated `base-scraper-with-browser.js` to validate `possibleResults` served in arrays and changed `possibleResults[LOGIN_RESULT.SUCCESS]` of `HapoalimScarper` to be array of the current possibilities. It doesn't supposed to affect non-array `possibleResults` members.
Besides that, the `fetchAccountData` function is now handling the different possibilities of the API URL by the current URL of the page entity.

I guess my coding style is not top notch, feel comfortable to change it as far as you want.

### DIFF
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -9,7 +9,7 @@ const VIEWPORT_WIDTH = 1024;
 const VIEWPORT_HEIGHT = 768;
 
 function getKeyByValue(object, value) {
-  return Object.keys(object).find(key => object[key] === value);
+  return Object.keys(object).find(key => object[key].indexOf(value) > -1);
 }
 
 function handleLoginResult(scraper, loginResult) {
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -9,7 +9,7 @@ const VIEWPORT_WIDTH = 1024;
 const VIEWPORT_HEIGHT = 768;
 
 function getKeyByValue(object, value) {
-  return Object.keys(object).find(key => object[key].indexOf(value) > -1);
+  return Object.keys(object).find(key => object[key].includes(value));
 }
 
 function handleLoginResult(scraper, loginResult) {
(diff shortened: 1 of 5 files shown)


src/scrapers/base-scraper-with-browser.js
```diff
@@ -9,7 +9,7 @@ const VIEWPORT_WIDTH = 1024;
 const VIEWPORT_HEIGHT = 768;
 
 function getKeyByValue(object, value) {
-  return Object.keys(object).find(key => object[key] === value);
```
eshaham: @chilikla I would rather use `includes(...)` as it's the modern way of doing `indexOf(...) > -1`

chilikla: I'm not sure your use case is possible, but maybe I didn't understand it.

As far as I understand, if `object[key]` (the success URL, as a string and not an array) equals `https://some.url.com/account`, and `value` (current URL) equals `https://some.url.com/account/error`, `object[key].includes(value)` (your suggestion is of course better) will return false. And this is still true if `object[key]` is an array.

If you agree with me, I will just commit the change to `object[key].includes(value)` and we are done.

eshaham: @chilikla AFAIK, `array.includes(value)` returns true only if the value EQUALS an item in the array.
so `['apple', 'orange'].includes('app')` will return false, whereas `'apple'.includes('app')` will return true.

chilikla:
I updated all the web based scarpers (as far as I understand, they're the Hapoalim, Discount and Leumi Card scarpers) with arrays in `getPossibleLoginResults`' `urls` object and updated `CONTRIBUTING.md` accordingly.

Do you think this is enough?

eshaham:
@chilikla can you fix the conflict?

chilikla:
@eshaham, done.

Issue #79, opened 2018-02-18
Multiple accounts for Bank Hapoalim scarper


chilikla:
Hi,

I'm looking into implementing multiple acounts for Bank Hapoalim scarper, but I couldn't get data for accounts other than the primary one.
Currently the scarper uses this url to get transactions data:
`${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`
And I couldn't succeeded using it with other accounts.
Maybe @sebikaplun  know how to crack it?

Besides that, I'm trying to implement account summary for the scraper. I guess with access to every account, the balance of every account will be located in the last transaction. But what about the credit limit of the account?
Again, @sebikaplun , do you have any idea, beside browser scarping?

chilikla:
OK, it turned out, I checked the above URL with account without transactions (not a checking account (עובר ושב)), and that's why the URL didn't work. It's supposed to work with every checking account.

As for the credit data, I fount this:
`${apiSiteUrl}/current-account/composite/balanceAndCreditLimit?accountId=${accountNumber}&view=details`
with the `currentBalance`, `currentAccountLimitsAmount` and `withdrawalBalance` properties.
This does the trick.

PR #82, opened 2018-02-23
Implement multiple accounts for Hapoalim scarper

chilikla:
Note: since for none-checking accounts the transactions API page doesn't exist, and in case of multiple accounts some of them will be of this kind, the scarper won't return error message when there are no transactions for the account, just an empty array.
I think it's reasonable, since there is no transactions API page also when there are no transactions in the wanted time frame, I think empty `txns` array is logical. Fix me or change it if you think differently.

### DIFF
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -27,34 +27,38 @@ function convertTransactions(txns) {
 async function fetchAccountData(page, options) {
   const apiSiteUrl = `${BASE_URL}/ServerServices`;
   const accountDataUrl = `${apiSiteUrl}/general/accounts`;
-  const accountInfo = await fetchGetWithinPage(page, accountDataUrl);
-  const accountNumber = `${accountInfo[0].bankNumber}-${accountInfo[0].branchNumber}-${accountInfo[0].accountNumber}`;
+  const accountsInfo = await fetchGetWithinPage(page, accountDataUrl);
 
-  const defaultStartMoment = moment().subtract(1, 'years').add(1, 'day');
-  const startDate = options.startDate || defaultStartMoment.toDate();
-  const startMoment = moment.max(defaultStartMoment, moment(startDate));
+  const accounts = [];
+  for (let accountIndex = 0; accountIndex < accountsInfo.length; accountIndex += 1) {
+    const accountNumber = `${accountsInfo[accountIndex].bankNumber}-${accountsInfo[accountIndex].branchNumber}-${accountsInfo[accountIndex].accountNumber}`;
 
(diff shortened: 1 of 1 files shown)


src/scrapers/hapoalim.js
```diff
@@ -27,34 +27,38 @@ function convertTransactions(txns) {
...
-  const startDateStr = startMoment.format(DATE_FORMAT);
-  const endDateStr = moment().format(DATE_FORMAT);
-  const txnsUrl = `${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`;
+    const defaultStartMoment = moment().subtract(1, 'years').add(1, 'day');
+    const startDate = options.startDate || defaultStartMoment.toDate();
+    const startMoment = moment.max(defaultStartMoment, moment(startDate));
 
-  const txnsResult = await fetchGetWithinPage(page, txnsUrl);
+    const startDateStr = startMoment.format(DATE_FORMAT);
+    const endDateStr = moment().format(DATE_FORMAT);
```
eshaham: @chilikla I believe `startDateStr` and `endDateStr` can be moved up, outside of the for loop block.

src/scrapers/hapoalim.js
```diff
@@ -27,34 +27,38 @@ function convertTransactions(txns) {
...
-      errorMessage: txnsResult.Error.MsgText,
-    };
+    let txns;
+    await fetchGetWithinPage(page, txnsUrl)
+      .then((txnsResult) => {
+        txns = convertTransactions(txnsResult.transactions);
+      })
+      .catch(() => {
+        txns = [];
+      });
```
eshaham: @chilikla should be re-written to something like:
```node
let txns;
try {
  txns = await fetchGetWithinPage(page, txnsUrl);
} catch () {
  txns = [];
}
txns = convertTransactions(txnsResult.transactions);
```

I assume you're creating an empty array here in case of exception, just because an account with no transactions will throw such an exception (as described in https://github.com/eshaham/israeli-bank-scrapers/issues/79)?

chilikla: Your suggestion has some mix up regarding the the used be variable there, `txnsResult`.
What do you think writing this:
```
let txns;
try {
  const txnsResult= await fetchGetWithinPage(page, txnsUrl);
  txns = convertTransactions(txnsResult.transactions);
} catch () {
  txns = [];
}
```

eshaham: @chilikla sure, that works too

Issue #83, opened 2018-03-02
Add an extra description property (Memo) for transactions


esakal:
@eshaham you created an essential project that is missing in the Israeli financial atmosphere! I really liked the way you scaffold the project and provided an infrastructure that is easy to follow and extend.

I'm working on a crawler for the Leumi Bank where they provide an extra field relevant to transfers into your account. This field describe the actual payee and it is an essential field for many transfer transactions like bit, הוראות קבע, salaries, internet wired transfers etc. Bank Hapoalim also exposes this information and I believe it is available for the others banking institutes as-well. 

I think the transaction should have an optional field named 'memo' of type string where we can provide this information.

I would appreciate your insights on that matter. 
Eran.

eshaham:
@esakal yes, it certainly makes sense, and `memo` as the field name sounds right.
Having said that, I would love to apply it as once for all banks, so it will be a joint effort.

If I understand correctly, you're working on the Leumi Bank scraper to be added here? If so, I would suggest separating the addition of the new field `memo` from your work on this scraper.

Btw, for Leumi Bank, @Urigo has mentioned he has something in place (unrelated to this repo), so it might be helpful for you to checkout his repo https://github.com/Urigo/leumi-leumicard-bank-data-scraper. Also, I think @ifeins started working on adding it here, so might be helpful to sync.

esakal:
@eshaham great lets do it. I commented 3 places in the PR for bank leumi #86  so you will be able to easily plug it back.

@ifeins, feel free to check my work in #86 . I added information about the scraper in the PR description. 

@Urigo did great work with his scrapers. Anyone who contribute to the open banking concept in Israel is a valued man. I chose a different strategy for the bank Leumi scraper since I prefer to use the 'wait' strategy against DOM anchors which doesn't care about network latency as opposed to 'pause for x seconds' strategy. I did like how he scraped the table and used the same approach with the puppeteer in the new scraper.

esakal:
@eshaham can you create a new branch for this feature?
Thanks!

eshaham:
@esakal done - called `add_memo_field`

eshaham:
@esakal pushed code to add `memo` field to Leumi Card and Isracard.
I also uncommented your code in Leumi scraper.
Discount doesn't seem to have any relevant field.
In Visa Cal there seem to be 2 relevant fields, but I don't have transactions with data in these fields, so couldn't figure out what to do. Maybe @nirgin could help?

[...]

PR #86, opened 2018-03-04
add new scraper for bank Leumi

esakal:
This PR contains a new scraper for bank Leumi. 

This PR also contains the following changes/additions:
- a playground area for testing/maintaining/fixing scrapers without leaving the project.
- extended login completed check - support RexExp to validate navigated url. This change was added since the urls of bank Leumi contain custom data per browser session (I verified this change with leumi-card, isracard, amex and leumi scrapers)
- `fillInput()` now cleans the content of the field before it types the new value
- `dropdownSelect()` was added to select value in input of type select

There are some things to be added later after this PR will be merged:
- multiple accounts in bank Leumi (I couldn't develop it since I only have one account)
- foreign account (I couldn't develop it since I don't have such an account)
- missing login completed check for password expired. I recently changed my password so it will take a while till I will see this scenario.

And some final issues to discuss about not related directly to this PR:
- export memo field was commented following eshaham/israeli-bank-scrapers#83
- imho there is a discrepancy between the type of the date as mentioned in file [contributing.md](https://github.com/eshaham/israeli-bank-scrapers/blob/master/CONTRIBUTING.md#overriding-fetchdata) and the actual values returned by all scrapers. The type should be date yet the scrapers return `moment(...).toISOString()` which is of type string. I did the same to be persist with other scrapers. 
- I think there is an issue with bank Hapoalim in the [following line](https://github.com/eshaham/israeli-bank-scrapers/blob/master/src/scrapers/hapoalim.js#L69). This is not the url I'm being navigated to so it fails after successful login. I didn't fix it since if I'm correct it doesn't relate to my PR. can you confirm it? 

@eshaham I didn't know if you want this PR to have single commit or if it doesn't matter since you squash the changes during the merge. If you need me to rebase in my branch let me know.


eshaham:
@esakal this is awesome!!
A few initial comments before I dive into this PR:
- It's a big PR which involves both the creation of the leumi scraper and the introduction of the "playground" concept. While I'm not entirely sure how I feel about the concept of playground (tend to like it, might need some adjustments), I know that it doesn't belong in this PR, and makes it much harder to review. Would you be able to move that into a different PR?
- I've just merged https://github.com/eshaham/israeli-bank-scrapers/pull/81, which also introduced a change in `getKeyByValue` function. In addition to that, that PR also addressed the issue of wrong expected URLs in Hapoalim, which was suspected to be some kind of experiment run by the bank - so that may be the answer to your comment about Hapoalim. Anyhow, could you check how it affects your changes to `getKeyByValue` and solve the conflict?
- You are absolutely right about the discrepancy between the type of the date as mentioned in file contributing.md and the actual values returned by all scrapers - will address it shortly
- Regarding squashing commits, I'm against it, as long as each commit stands on it's own - so need need to change anything 😄 
- One final comment (ANAL COMMENT ALERT!!!!!), and that's already my own ADHD, would be great if you could reorder the introduction of Leumi in different files, like `README` and `definitions.js`, and put it after Hapoalim - I like to keep scrapers according to importance - first banks, then credit cards. I again apologize for the analism of my comment 🥇

esakal:
closing this PR in favor of #90 and #91

Issue #76, opened 2018-02-16
Account summary alongside the transactions data


chilikla:
My use case, for now, of using this library is getting all of my finance information in one place. For this matter, transactions data is less important for me, but accounts summaries are more useful. e.g. getting the current balance of my bank account(s), or knowing the my expected credit card(s) charge for this month.
Is this something this project willing to include?
I already wrote a code for myself for Bank Hapoalim and Leumi Card with new properties of 'balance' (in the case of the bank) and 'charge' (in the case of the credit card), in the returned object (alongside the success and txns properties).

In a related manner, what about investing data?
I have an Intergemel account, and I'm finding it useful to get data of "what my money did today" for it. I can contribute code for handling Intergemel accounts, with basic information about the type of account, its number, its nick name and its current balance of each of the customer accounts.

eshaham:
Hey @chilikla 😄
I think that adding account summaries would be a great add-on to existing data we scrape.
can you suggest the signature changes for the existing scrape result?
I wonder if we could still maintain the same signature for both banks and credit cards when it comes to summaries. Something tells me it's going to be difficult.
We would need to add these to all scrapers of course, so it will be a joint effort.

As for Intergemel, I'm thinking it may not fit here. I've actually started another project and started scraping insurance/pension companies, but it is in a very initial stage:
https://github.com/eshaham/israeli-insurance-scrapers
see clalbit branch.

chilikla:
Hi @eshaham :)
I don't really understand what do you mean by maintaining same signature. I found out you used signatures for marking accounts, but I don't understand the connection of this to the issue.
I can only tell about the accounts I have access to. Every transaction in Bank Hapoalim is returning the account balance after the transaction has been made, so it's quite easy to know the current balance of the account. The problem is that Hapoalim scraper doesn't support multiple accounts (is it planned?), but after it will implemented, it will be just there.
As for Leumi Card, this is actually problematic in some way. The website is providing the next month charge in the homepage, but it's the charge for all of your cards combined. In the transactions page, all of charges are detailed, but they are separated to ILS charges and foreign currency charges, and there are also postponed charges which makes the calculation more difficult. I chose to obtain the whole next month charge, and I couldn't find an easy way to get the month charges per account, but maybe this is the route we will choose to walk in and we should pursuit it, even if it somehow complicated.

eshaham:
@chilikla sorry for not making it clearer.
[...]

PR #103, opened 2018-04-01
Add account summary property to Hapoalim scraper

chilikla:
As detailed in issue #76, the scraper now adds a summary property to every account the scraper returns.
The summary structured like this:
```
{
  ...
  summary: {
    balance: double,
    creditLimit: double,
    creditUtilization: double,
    balanceCurrency: 'ILS',
  }
  ..
}
```
(I created faulty PR before, this is a fix.)

### DIFF
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -24,6 +24,15 @@ function convertTransactions(txns) {
   });
 }
 
+function fetchSummary(balanceCreditResult) {
+  return {
+    balance: parseFloat(parseFloat(balanceCreditResult.currentBalance).toFixed(2)),
+    creditLimit: parseFloat(parseFloat(balanceCreditResult.currentAccountLimitsAmount).toFixed(2)),
+    creditUtilization: parseFloat(parseFloat(balanceCreditResult.withdrawalBalance).toFixed(2)),
+    balanceCurrency: 'ILS',
+  };
+}
+
 async function fetchAccountData(page, options) {
   const currentUrl = await getCurrentUrl(page, true);
   const subfolder = (currentUrl.includes('portalserver')) ? 'portalserver' : 'ssb';
@@ -42,8 +51,16 @@ async function fetchAccountData(page, options) {
   for (let accountIndex = 0; accountIndex < accountsInfo.length; accountIndex += 1) {
     const accountNumber = `${accountsInfo[accountIndex].bankNumber}-${accountsInfo[accountIndex].branchNumber}-${accountsInfo[accountIndex].accountNumber}`;
 
-    const txnsUrl = `${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`;
+    const balanceCreditUrl = `${apiSiteUrl}/current-account/composite/balanceAndCreditLimit?accountId=${accountNumber}&view=details`;
+    let summary;
+    try {
+      const balanceCreditResult = await fetchGetWithinPage(page, balanceCreditUrl);
+      summary = fetchSummary(balanceCreditResult);
+    } catch (e) {
+      summary = {};
+    }
 
+    const txnsUrl = `${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`;
     let txns;
     try {
       const txnsResult = await fetchGetWithinPage(page, txnsUrl);
@@ -54,6 +71,7 @@ async function fetchAccountData(page, options) {
 
     accounts.push({
       accountNumber,
+      summary,
       txns,
     });
   }
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -33,9 +33,18 @@ function convertToSummary(balanceCreditResult) {
   };
 }
 
+function getSubFolder(currentUrl) {
+  if (currentUrl.includes('portalserver')) {
+    return 'portalserver';
+  } else if (currentUrl.includes('ng-portals')) {
+    return 'ServerServices';
+  }
+  return 'ssb';
+}
+
 async function fetchAccountData(page, options) {
   const currentUrl = await getCurrentUrl(page, true);
-  const subfolder = (currentUrl.includes('portalserver')) ? 'portalserver' : 'ssb';
+  const subfolder = getSubFolder(currentUrl);
   const apiSiteUrl = `${BASE_URL}/${subfolder}`;
   const accountDataUrl = `${BASE_URL}/ServerServices/general/accounts`;
   const accountsInfo = await fetchGetWithinPage(page, accountDataUrl);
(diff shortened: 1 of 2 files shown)


src/scrapers/hapoalim.js
```diff
@@ -24,6 +24,15 @@ function convertTransactions(txns) {
   });
 }
 
+function fetchSummary(balanceCreditResult) {
+  return {
+    balance: parseFloat(parseFloat(balanceCreditResult.currentBalance).toFixed(2)),
+    creditLimit: parseFloat(parseFloat(balanceCreditResult.currentAccountLimitsAmount).toFixed(2)),
+    creditUtilization: parseFloat(parseFloat(balanceCreditResult.withdrawalBalance).toFixed(2)),
```
eshaham: @chilikla are you sure `balanceCreditResult.withdrawalBalance` is the right prop to use for `creditUtilization`?
I don't have an Hapoalim account, so I have no way of seeing what it represents.

src/scrapers/hapoalim.js
```diff
@@ -42,8 +51,16 @@ async function fetchAccountData(page, options) {
   for (let accountIndex = 0; accountIndex < accountsInfo.length; accountIndex += 1) {
     const accountNumber = `${accountsInfo[accountIndex].bankNumber}-${accountsInfo[accountIndex].branchNumber}-${accountsInfo[accountIndex].accountNumber}`;
 
-    const txnsUrl = `${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`;
+    const balanceCreditUrl = `${apiSiteUrl}/current-account/composite/balanceAndCreditLimit?accountId=${accountNumber}&view=details`;
+    let summary;
+    try {
+      const balanceCreditResult = await fetchGetWithinPage(page, balanceCreditUrl);
+      summary = fetchSummary(balanceCreditResult);
+    } catch (e) {
```
eshaham: @chilikla why are you swallowing this exception?
If this call fails, the exception should probably propagate all the way up.

src/scrapers/hapoalim.js
```diff
@@ -24,6 +24,15 @@ function convertTransactions(txns) {
   });
 }
 
+function fetchSummary(balanceCreditResult) {
```
eshaham: @chilikla perhaps rename to `convertToSummary`? This function doesn't "fetch" anything.

src/scrapers/hapoalim.js
chilikla: @eshaham you think if there is no summary property, this account shouldn't be submitted at all?

src/scrapers/hapoalim.js
chilikla: @eshaham it's a good question. This property is saying how much left for you to withdraw. So if you are in overdraft, this is `creditUtilization` per se. But if you have plus in your bank account, it's  a sum of your `balance` and `creditLimit`.
What do you think will be preferred in case of plus - returning 0 (since the utilization is zero right now) or the current sum?

eshaham: @chilikla in that case, in case of an overdraft, shouldn't `creditUtilization` be `-balance`? I know we said something different in #76, but it seems to me this is the right thing in case of a bank, isn't it?
In case of a positive balance, I think it should be 0.

src/scrapers/hapoalim.js
eshaham: @chilikla we don't expect this call to fail on a regular basis, and once account summary will be something the users of this module expect, the scraper should crash in case it cannot get this expected data.

src/scrapers/hapoalim.js
chilikla: @eshaham there is `balanceCreditResult.creditLimitUtilizationAmount`, which indeed has 0 for me right now (since I'm not in overdraft). I think it's logical to use it. I just don't know what will its value in case of overdraft. What do you think?

src/scrapers/hapoalim.js
chilikla: @eshaham understood.

src/scrapers/hapoalim.js
eshaham: @chilikla hmmm, sounds like it has the right name...
We can either try to use it, or do `-balance`.
Your call 😄

chilikla: @eshaham I think `creditLimitUtilizationAmount` is better, since I'm not sure `-balance` is the right value for utilization but how much left to withdraw from the account, or `creditLimit + balance` (in case of negative value for `balance`).

eshaham: @chilikla credit utilization should be how much you've used off your credit limit, so it should probably be something like `Math.max(-balance, creditLimit)`. I wonder what happens in the bank when you go over your credit limit... Anyhow, I guess `creditLimitUtilizationAmount` has the right name, so we should probably go with that... 😄

src/scrapers/hapoalim.js
eshaham: @chilikla saw what you did in 79bef319cd557a91e9b3e5b3987729d86ae13076.
What I actually meant was, get rid of the `try..catch` clause altogether - you're still swallowing the exception ATM.

src/scrapers/hapoalim.js
```diff
@@ -55,25 +55,25 @@ async function fetchAccountData(page, options) {
...
+      summary = convertToSummary(balanceCreditResult);
+
+      const txnsUrl = `${apiSiteUrl}/current-account/transactions?accountId=${accountNumber}&numItemsPerPage=150&retrievalEndDate=${endDateStr}&retrievalStartDate=${startDateStr}&sortCode=1`;
+      let txns;
+      try {
+        const txnsResult = await fetchGetWithinPage(page, txnsUrl);
+        txns = convertTransactions(txnsResult.transactions);
+      } catch (err) {
+        txns = [];
+      }
```
eshaham: @chilikla same here for `try..catch` - we don't need it as we're covered by the base scrapers in case something fails.

chilikla: @eshaham 
Sorry for the absence lately, but I returned :relaxed:

Regarding to getting rid of the two `try... catch` blocks, I understand your point, but there are 2 scenarios which it will be problematic in that case:

- In case of moving accounts between branches, Hapoalim is keeping the old accounts accessible through their website (along with the new active ones), accessible but empty. Their particular API pages are empty and therefore failing the summary fetch action, even though there is no actual problem since the active accounts already fetched. Using `try... catch` mechanism will return the real accounts data and won't fail the whole scarping.

- Accounts with not transactions will fail the scraper as well, in the transactions fetch actions. And I'm not sure it won't be better to just return empty array of transactions, since there is no problem with the account per se, it just silent.

What do you think?

eshaham: > Their particular API pages are empty and therefore failing the summary fetch action, even though there is no actual problem since the active accounts already fetched.

Got it. Then while I usually avoid comments in code, this is a classic place to add a comment explaining why the `try...catch` is needed here.

> Accounts with not transactions will fail the scraper as well, in the transactions fetch actions

Assuming this is indeed the case for accounts without transactions, then same response as above 😄
Once you add a comment I think we can merge this.

chilikla:
@eshaham, I added the comments, hope in the wanted format.
I also had to add #115 changes for the login to work.

PR #107, opened 2018-04-02
add status property to bank leumi scraper

esakal:
scrape status of transactions.

**Pending transactions list as shown in the bank site**
scrape-status

**Output**
image

### DIFF
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -167,9 +167,11 @@ async function fetchTransactionsForAccount(page, startDate) {
 
   const accountNumber = selectedSnifAccount.replace('/', '_');
 
+  const pendingTxns = await extractPendingTransactionsFromPage(page);
+  const completedTxns = await extractCompletedTransactionsFromPage(page);
   const txns = [
-    ...await extractPendingTransactionsFromPage(page),
-    ...await extractCompletedTransactionsFromPage(page),
+    ...pendingTxns,
+    ...completedTxns,
   ];
 
   return {
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -154,7 +154,7 @@ async function fetchTransactionsForAccount(page, startDate) {
   await fillInput(
     page,
     'input#dtFromDate_textBox',
-    startDate.format('DD/MM/YY'),
+    startDate.format(DATE_FORMAT),
   );
   await clickButton(page, 'input#btnDisplayDates');
   await waitForNavigation(page);
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -52,30 +52,15 @@ function convertTransactions(txns) {
       originalAmount: amount,
       originalCurrency: SHEKEL_CURRENCY,
       chargedAmount: amount,
+      status: txn.status,
       description: txn.description,
       /* memo: txn.memo, TODO add this line to export transaction memo */
     };
   });
 }
 
-async function fetchTransactionsForAccount(page, startDate) {
-  await dropdownSelect(page, 'select#ddlTransactionPeriod', '004');
-  await waitUntilElementFound(page, 'select#ddlTransactionPeriod');
-  await fillInput(
-    page,
-    'input#dtFromDate_textBox',
-    startDate.format('DD/MM/YY'),
-  );
-  await clickButton(page, 'input#btnDisplayDates');
-  await waitForNavigation(page);
-  await waitUntilElementFound(page, 'table#WorkSpaceBox table#ctlActivityTable');
-  await clickButton(page, 'a#lnkCtlExpandAll');
-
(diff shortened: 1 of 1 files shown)


src/scrapers/leumi.js
```diff
@@ -119,6 +103,75 @@ async function fetchTransactionsForAccount(page, startDate) {
...
+
+  const selectedSnifAccount = await page.$eval('#ddlAccounts_m_ddl option[selected="selected"]', (option) => {
+    return option.innerText;
+  });
+
+  const accountNumber = selectedSnifAccount.replace('/', '_');
+
+  const txns = [
+    ...await extractPendingTransactionsFromPage(page),
+    ...await extractCompletedTransactionsFromPage(page),
```
eshaham: @esakal can you extract these calls into separate variables?

src/scrapers/leumi.js
```diff
@@ -119,6 +103,75 @@ async function fetchTransactionsForAccount(page, startDate) {
...
+      txns.push(changedTransaction);
+    } else if (element.classList.includes('Colume5Width')) {
+      const changedTransaction = txns.pop();
+      changedTransaction.credit = element.innerText;
+      txns.push(changedTransaction);
+    } else if (element.classList.includes('Colume6Width')) {
+      const changedTransaction = txns.pop();
+      changedTransaction.balance = element.innerText;
+      txns.push(changedTransaction);
+    }
```
eshaham: @esakal what's with all the pushes and pops?
Wouldn't it be much simpler to iterate over all `tr`s instead of all `td`s?
That way you will know when the row ends and could access a specific cell by index.

src/scrapers/leumi.js
```diff
@@ -119,6 +103,75 @@ async function fetchTransactionsForAccount(page, startDate) {
...
+
+async function extractPendingTransactionsFromPage(page) {
+  const txns = [];
+
+  const tdsValues = await page.$$eval('#WorkSpaceBox #trTodayActivityNapaTableUpper tr td', (tds) => {
+    return tds.map(td =>
+      ({
+        classList: td.getAttribute('class'),
+        innerText: td.innerText,
+      }));
```
eshaham: @esakal weird syntax, I believe you did that because eslint was shouting at you. 
What it meant was probably:
```node
return tds.map((td) => {
  classList: td.getAttribute('class'),
  innerText: td.innerText,
});
```

src/scrapers/leumi.js
```diff
@@ -119,6 +103,75 @@ async function fetchTransactionsForAccount(page, startDate) {
...
+  return txns;
+}
+
+async function fetchTransactionsForAccount(page, startDate) {
+  await dropdownSelect(page, 'select#ddlTransactionPeriod', '004');
+  await waitUntilElementFound(page, 'select#ddlTransactionPeriod');
+  await fillInput(
+    page,
+    'input#dtFromDate_textBox',
+    startDate.format('DD/MM/YY'),
```
eshaham: @esakal not new code, but would be a good time to extract the date format into a constant

src/scrapers/leumi.js
esakal: You suggestion will not work, see image below. It is not related to eslint. You must wrap the resulted object with `(` otherwise it will treat `{` as an opening of a function scope and not an opening of json object notation

image

src/scrapers/leumi.js
esakal: @Urigo did it in his library at [line 206](https://github.com/Urigo/leumi-leumicard-bank-data-scraper/blob/master/src/scraping/leumi.js#L206). Since it works well and seems solid I think we should stick with what he did.

src/scrapers/leumi.js
esakal: I'm not sure, I'm a fan of constants where the value is used in multiple places. The date format value is specific to this function logic so I don't think we should move it outside . wdyt?

src/scrapers/leumi.js
esakal: done

esakal:
@eshaham did you have a chance to see my responses?

src/scrapers/leumi.js
eshaham: @esakal 😞still don't like it, but if you insist...

src/scrapers/leumi.js
eshaham: @esakal my bad, missed the fact you were returning an object.

src/scrapers/leumi.js
eshaham: @esakal I usually try to extract important variables into constants even if they're only used once, so they appear loud and clear at the top of files. Again, see your point, so not going to insist over that either.

esakal:
@eshaham Morning :) 
Out of the 2 pending comments, I fixed the one with the date format written not using constant. I prefer not to touch the second one. If you can merge it that would be great :)

eshaham:
@esakal 👍

PR #108, opened 2018-04-05
scrape memo for bank leumi

esakal:
@eshaham please review the following. Since I already supported scraping memos when I added the scraper for Bank Leumi all I needed was to uncomment the following line.

## Output
image


esakal:
the last commit is needed to remove the new line character that was added to the summary value 

image

eshaham:
@esakal I think you're confusing the `account_summary` branch with `add_memo_field` branch.
I already uncommented the memo field in the `add_memo_field` branch, see this commit: cc47373431e3d5292bad6425f2038a5a22067047.

The `account_summary` branch is used to answer #76

PR #120, opened 2018-05-21
Hapoalim bugfix

matanelgabsi:
Hapoalim started using XSRF token for transactions (and pageuuid etc.)


eshaham:
@matanelgabsi can you update the PR comment with an explanation for what doesn't work at the moment?

PR #121, opened 2018-05-21
visa-cal: handle txn type 72

esakal:
# Motivation
support transactions of type "שירותים"
image

# Current behavior
scraping fails due to invalid transaction type 72

# New behavior
transaction type 72 is considered as normal transaction


eshaham:
@esakal hmmm, that's interesting.
Maybe it's time to introduce a new transaction type for card/bank fees?
It would definitely be beneficial for users of this library in some use cases.
wdyt?

esakal:
@eshaham done! since the type description is `שירותים` i called it `SERVICES_TYPE_CODE`

PR #122, opened 2018-05-21
visa-cal: handle txn type 67

esakal:
# Motivation
support transactions of type "דמי חבר"
image

# Current behavior
scraping fails due to invalid transaction type 67

# New behavior
transaction type 67 is considered as normal transaction


src/scrapers/visa-cal.js
```diff
@@ -67,6 +69,7 @@ function convertTransactionType(txnType) {
     case WITHDRAWAL_TYPE_CODE:
     case WITHDRAWAL_TYPE_CODE_2:
     case REFUND_TYPE_CODE_2:
+    case NORMAL_TYPE_CODE_3:
```
esakal: @eshaham since there is only one type representing installments and all the others represent normal transactions, we can avoid missing types support in the future by settings normal as default. Let me know if you want me to do it here, if so it will close #121 as well

eshaham:
@esakal again, this might as well be yet another transaction type for membership fees (probably a different one from the one I mentioned in #121 - wdyt?

src/scrapers/visa-cal.js
eshaham: @esakal I actually like this process of approving txn types one by one, as it makes me think about each one, see my comment to the PR.

eshaham:
@esakal Same comment as I made in https://github.com/eshaham/israeli-bank-scrapers/pull/121#issuecomment-396030991

esakal:
@eshaham done! since the type description is `דמי חבר` I called it `MEMBERSHIP_FEE_TYPE_CODE`

PR #127, opened 2018-06-01
visa-cal: handle txn type 71

esakal:
@eshaham you will have a minor conflict with #121  and #122 but it should be quite trivial to merge.


eshaham:
@esakal what's does this new type stand for?

esakal:
@eshaham done! since the type description is `החזר שירותים` I called it `SERVICES_REFUND_TYPE_CODE`

Issue #112, opened 2018-04-10
Leumi card scrapers fails to report errors during scraping


esakal:
https://github.com/eshaham/israeli-bank-scrapers/blob/ecdf9777f23280140370ffc79729d96449c8bddb/src/scrapers/leumi-card.js#L234-L236

## Bug Overview
the scraper currently ignore failures for line 234 `goto` command, resulting with "valid" scraping of 0 transactions instead of failing the process.

## Reproduce Scenario
change the value of the url to invalid value, for example `path: 'Registred/Transactions/invalid-page-name.aspx`

https://github.com/eshaham/israeli-bank-scrapers/blob/ecdf9777f23280140370ffc79729d96449c8bddb/src/scrapers/leumi-card.js#L39

## Expected result
an error should be thrown, stopping the process and providing informative information about the error. 

## Actual result
the process complete with status `success` as shown here
![image](https://user-images.githubusercontent.com/4751797/38563870-96c56810-3ce6-11e8-9fa4-eae84fe178c1.png)

and the resulted csv file content is:
![image](https://user-images.githubusercontent.com/4751797/38563894-a2b6d94c-3ce6-11e8-92b9-e3ab6a86c02d.png)

PR #129, opened 2018-06-10
throw error if navigation status not OK

eshaham:
This PR implements a check to see the status of each navigation. If not OK, a.k.a 200, it will throw an error.

related to #112 - but solves a different problem


esakal:
@eshaham I liked your addition. It will probably solve general navigation issues. You motivated me to find the cause for #112. I think that I found how to handle that specific scenario in #130.

PR #133, opened 2018-06-25
Fix leumi html table selector

shuki770:
Chenge leumi selector id from "trTodayActivityNapaTableUpper" to "ctlActivityTable"

### DIFF
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -112,7 +112,7 @@ async function extractCompletedTransactionsFromPage(page) {
 async function extractPendingTransactionsFromPage(page) {
   const txns = [];
 
-  const tdsValues = await page.$$eval('#WorkSpaceBox #trTodayActivityNapaTableUpper tr td', (tds) => {
+  const tdsValues = await page.$$eval('#WorkSpaceBox #ctlActivityTable tr td', (tds) => {
     return tds.map(td => ({
       classList: td.getAttribute('class'),
       innerText: td.innerText,


eshaham:
@shuki770 since I don't have a Leumi account, can you please explain (preferably edit the PR description) what the behavior before and after this change?

esakal:
Hi @shuki770,
There are two types of tables in Leumi site:
- `ctlActivityTable` - is the default table that holds all your _regular_ transactions and can exists multiple times in the same page.
- `trTodayActivityNapaTableUpper` - is a unique table that holds all your _pending_ transactions and usually not exists on the page.
IMO, changing the  selector to `ctlActivityTable` will result with duplicate transactions and also you will not mark correctly `pending` transactions. wdyt?

Reading the error you provided and reviewing the code I think the error message poorly describe what is actually happening. In my machine I'm not getting this error and it works as expected by returning empty array, maybe we have different puppeteer versions? which one are you using?

If this error still happens in your machine when you are using the original scraper code, please try the following:
```
const tdsValues = await page.$$eval('#WorkSpaceBox #trTodayActivityNapaTableUpper tr td', (tds) => {
    return tds ? tds.map(td =>
      ({
        classList: td.getAttribute('class'),
        innerText: td.innerText,
      })) : [];
  });
```

Eran.

esakal:
@shuki770 I find the reason for this error and it relates to a breaking change that was done by mistake in `puppeteer` (the folder that does the actual scraping). 

We are using puppeteer v1.5.0 as part of [israeli-bank-scrapers@0.6.0](https://github.com/eshaham/israeli-bank-scrapers/commit/1393ce5f04ce82f438087abad78c5075ce948c9a)

23 days ago they released version 1.5.0 where they changed the logic of `$$eval` and as a side effect it started failing when the selector is missing. this was not their original intention and they fix it for next version [here](https://github.com/GoogleChrome/puppeteer/pull/2740).

I will merge a fix for that issue soon so we will not need to wait for their release.

shuki770:
Thank you! @esakal 

If the trTodayActivityNapaTableUpper table does not usually exist. It is not better to use the ctlActivityTable table ?

esakal:
This library scrape transactions that are either completed or pending. each selector is used to scrape different transaction status. `ctlActivityTable` selector is much more common then `trTodayActivityNapaTableUpper` but you need both of them. see the mapping attached below.

We already have a fix for that in #135, it will be merged soon.

image

PR #134, opened 2018-06-30
Hapoalim change password

esakal:
# Overview
Bank hapoalim has another url that is used when password expires. 

## Current behavior
When getting the following page, the scraper return with status `undefined`
image

## Modified behavior
The hapoalim scraper now recognize this page url.

# Technical
## Why using Regexp instead of string comparison
the url of this page is as followed: `https://login.bankhapoalim.co.il/ABOUTTOEXPIRE/START?flow=ABOUTTOEXPIRE&state=START&expiredDate=11072018`. The last property named `expiredDate` is dynamic and changes per user. So we need to be able to compare using regexp

## Why changing the base scraper functionality 
property `loginOptions.possibleResults` structure is hashmap and hapoalim already have a string comparison value for key `LOGIN_RESULT.CHANGE_PASSWORD`. I needed to modify the logic to allow passing in the array value both string and regexp.

### DIFF
--- a/src/scrapers/hapoalim.js
+++ b/src/scrapers/hapoalim.js
@@ -127,7 +127,10 @@ function getPossibleLoginResults() {
   const urls = {};
   urls[LOGIN_RESULT.SUCCESS] = [`${BASE_URL}/portalserver/HomePage`, `${BASE_URL}/ng-portals-bt/rb/he/homepage`, `${BASE_URL}/ng-portals/rb/he/homepage`];
   urls[LOGIN_RESULT.INVALID_PASSWORD] = [`${BASE_URL}/AUTHENTICATE/LOGON?flow=AUTHENTICATE&state=LOGON&errorcode=1.6&callme=false`];
-  urls[LOGIN_RESULT.CHANGE_PASSWORD] = [`${BASE_URL}/MCP/START?flow=MCP&state=START&expiredDate=null`];
+  urls[LOGIN_RESULT.CHANGE_PASSWORD] = [
+    `${BASE_URL}/MCP/START?flow=MCP&state=START&expiredDate=null`,
+    /[.]co[.]il\/ABOUTTOEXPIRE\/START/i,
+  ];
   return urls;
 }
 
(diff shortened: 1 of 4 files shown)


src/scrapers/hapoalim.js
```diff
@@ -127,7 +127,10 @@ function getPossibleLoginResults() {
   const urls = {};
   urls[LOGIN_RESULT.SUCCESS] = [`${BASE_URL}/portalserver/HomePage`, `${BASE_URL}/ng-portals-bt/rb/he/homepage`, `${BASE_URL}/ng-portals/rb/he/homepage`];
   urls[LOGIN_RESULT.INVALID_PASSWORD] = [`${BASE_URL}/AUTHENTICATE/LOGON?flow=AUTHENTICATE&state=LOGON&errorcode=1.6&callme=false`];
-  urls[LOGIN_RESULT.CHANGE_PASSWORD] = [`${BASE_URL}/MCP/START?flow=MCP&state=START&expiredDate=null`];
+  urls[LOGIN_RESULT.CHANGE_PASSWORD] = [
+    `${BASE_URL}/MCP/START?flow=MCP&state=START&expiredDate=null`,
+    /[.]co[.]il\/ABOUTTOEXPIRE\/START/i,
```
eshaham: @esakal if you choose to include the `co.il` part in the regex expression, maybe you should include the `BASE_URL`? I'm also ok with dropping the domain part altogether.

esakal: @eshaham I knew it will come up :) jslint has been giving me a hard time. In order to prefix the expression with `BASE_URL` I needed to use `RegExp` and thus escape the argument. 

By doing so, jslint error about `no-useless-escape`.
image

So I tried to concat them and got `prefer-template` error from jslint instead
image

Since you don't like bypassing jslint rules I tried to find a way that will satisfy jslint and this is what I come up with :) Are you familiar with an other approach I can try here which will be acceptable by jslint?

eshaham:
@esakal 🚢 it

PR #135, opened 2018-06-30
fix undesired behavior of pupeteer for page.$$eval started since pupp…

esakal:
We are using puppeteer v1.5.0 as part of [israeli-bank-scrapers@0.6.0](https://github.com/eshaham/israeli-bank-scrapers/commit/1393ce5f04ce82f438087abad78c5075ce948c9a)

23 days ago they released version 1.5.0 where they changed the logic of `$$eval` and as a side effect it started failing when the selector is missing. this was not their original intention and they fix it for next version [here](https://github.com/GoogleChrome/puppeteer/pull/2740).

closes #133

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,9 +7,16 @@
       "error",
       "ForInStatement",
       "LabeledStatement",
-      "WithStatement",
+      "WithStatement"
     ],
-    "operator-linebreak": ["error", "after"]
+    "operator-linebreak": ["error", "after"],
+    "max-len": ["error", 100, 2, {
+      "ignoreUrls": true,
+      "ignoreComments": true,
+      "ignoreRegExpLiterals": true,
+      "ignoreStrings": true,
+      "ignoreTemplateLiterals": true
+    }]
   },
   "globals": {
     "document": true,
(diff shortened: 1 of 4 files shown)


eshaham:
> I can do that but since we assume it will be fixed as part of puppeteer next version I'm not sure if it is needed. Still I don't mind doing it if you think it is beneficial. wdyt?

I think we should extract stuff to shared functions whenever we get the chance. That way, the next time puppeteer change anything in their signatures / behaviors, we will only need to fix in one place.
I started doing that with some functions, but it is time consuming... :-) this code change is a good opportunity to extract `eval` function.

> I'm also a fan of TODOs but jslint has been giving me a hard time here as well with rule max-len rule. 

🤦‍♂️ 

> We can however set max-lenoption skip-comments to remove this limitation specific for comments.

Let's do that 😄

.eslintrc
```diff
@@ -7,9 +7,16 @@
       "error",
       "ForInStatement",
       "LabeledStatement",
-      "WithStatement",
+      "WithStatement"
     ],
-    "operator-linebreak": ["error", "after"]
+    "operator-linebreak": ["error", "after"],
```
esakal: I needed to copy the default rule configuration from air-bnb to disable only `ignoreComments` option.

esakal:
@eshaham please review when possible

.eslintrc
eshaham: @esakal 👍

PR #145, opened 2018-07-16
optional external browser instance 

### LINKED ISSUE #142: launch puppeteer browser outside the lib
I found it hard to control the launch command of puppeteer. 
I would be super nice If I could pass `browser` object to `createScraper(option)` function.

First of all: how do you feel about it❓ 
as I see it there is couple of options: 
## option 1 : 
```
  const options = {
    companyId: 'isracard', 
    startDate: '01-06-2018', 
    combineInstallments: false,
    verbose: true, // include more debug info about in the output
    browser: browser,
  };
``` 
where browser is `const browser = await puppeteer.launch(SomeParams);`

## option2: ```createScraper(options,maybeBrowser)```

both options will need to something like this:  
```
if(browser !=null) {
  this.browser = browser;
}else {
  this.browser  = await puppeteer.launch();
}
```

anything else?

lielran:
closes #142

### DIFF
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -63,7 +63,7 @@ class BaseScraperWithBrowser extends BaseScraper {
     if (this.options.verbose) {
       env = Object.assign({ DEBUG: '*' }, process.env);
     }
-    if (this.options.browser != null) {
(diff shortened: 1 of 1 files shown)


lielran:
@esakal change to `!==`

lielran:
@esakal  fixed

PR #150, opened 2018-07-21
Leumi Scraper - fix expand link id.

ifeins:
The expand all transactions link id has changed which caused the scraper to break.


esakal:
@ifeins thanks! I will merge it in a minute. As it might take a while until it will deploy to npmjs, I just want to make sure that you are familiar with `npm link` which will allow you to use this library locally against branch master. I know you probably know this thing but I just want to be sure.

PR #154, opened 2018-08-15
Visa Cal - If bank account has no active cards, the scrape fails

ErezSavir:
The scrape fails if you have 2 or more bank accounts and all cards are inactive resulting in a generic error

### DIFF
--- a/src/scrapers/visa-cal.js
+++ b/src/scrapers/visa-cal.js
@@ -190,29 +190,28 @@ async function getTransactionsForAllAccounts(authHeader, startMoment, options) {
     for (let i = 0; i < banksResponse.BankAccounts.length; i += 1) {
       const bank = banksResponse.BankAccounts[i];
       const bankDebits = await getBankDebits(authHeader, bank.AccountID);
-      if (bank.Cards.every(function (element){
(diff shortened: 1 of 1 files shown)


src/scrapers/visa-cal.js
```diff
@@ -190,25 +190,28 @@ async function getTransactionsForAllAccounts(authHeader, startMoment, options) {
...
-          if (rawTxns) {
-            let txns = convertTransactions(rawTxns);
-            txns = prepareTransactions(txns, startMoment, options.combineInstallments);
-            const result = {
-              accountNumber: bank.Cards[j].LastFourDigits,
-              txns,
-            };
-            accounts.push(result);
+      //Check that the bank has an active card to scrape
+      if (bank.Cards.some(_ => _.IsEffectiveInd)) {
```
eshaham: @ErezSavir looks much better :-)
My only comment now is about the usage of underscore `_`.
It's much nicer to see a relevant variable here. Besides, it's confusing due to the usage of underscore for lodash related operations.
So would be great if you could convert it to `bank.Cards.some(card => card.IsEffectiveInd)`.

PR #156, opened 2018-09-05
bank-leumi: scrape foreign USD account

esakal:


esakal:
@eshaham @asfktz @lielran if one of you have an account in bank Leumi **without foreign (sub)account**, can you please execute this code and verify that it doesn't crash? I want to make sure that it works as expected.

Also, I added support for USD currency only as this is what I have in my account, It is super easy to support other currencies by adding 2 lines in the following [switch...case](https://github.com/eshaham/israeli-bank-scrapers/blob/charged-currency-bank-leumi/src/scrapers/leumi.js#L242). If one of you has other type of currencies just send me the value that is shown in the table for that currency. As an example for USD the value is  `דולר ארה''ב`

thanks :)

eshaham:
@esakal I don't have an active Leumi account to verify

eshaham:
@esakal it took me some time to understand why I am confused by this PR, but then I realized.
I don't think this PR is related to `charged-currency` branch / initiative. It's not adding support for transactions charged in different currencies in the scraped account, but rather adds support for other accounts in different currencies. While it feels like something we should consider supporting (though we could argue whether it's more important to add support for "loan accounts" for example), I don't think this change belongs in this initiative.
As far as I see it, since that in regular bank account there's no support for transactions in other currencies, `charged-currency` should only add changes to credit card scrapers.
wdyt?

esakal:
@eshaham I understand your point but I think they are coupled together and I'm not sure if can provide one without the other. Well obviously we can but it makes this feature useless when importing into budgeting systems like YNAB. Let me explain, the only reason/scenario for a credit card to charge you in foreign currency afaik is if that card is linked to foreign account in your bank. Which means that if you want to keep your budget complete you must also track foreign accounts, otherwise any reasonable budgeting system will treat those transactions has outcomes (instead of transfers) and will affect the budget integrity. 

So I agree that it might not be relevant to charged currency directly but it is needed for the integrity of the budget tracking as explained above, and in addition it does use the `chargedCurrency` to set a value different then `ILS`. 

Does this make sense to you?

esakal:
following @eshaham comment, closes this PR in favor of #159

PR #157, opened 2018-09-05
leumi-card: scrape charged currency transaction property 

esakal:


src/scrapers/leumi-card.js
```diff
@@ -151,10 +150,62 @@ async function getAccountNumber(page, cardIndex) {
...
+  if (typeof amountStr === 'undefined' || amountStr === null ||
+    (typeof amountStr === 'string' && amountStr.trim().length === 0)) {
+    return null;
+  }
+
+  const formattedAmount = amountStr
+    .replace(SHEKEL_CURRENCY_SYMBOL, '')
+    .replace(DOLLAR_CURRENCY_SYMBOL, '')
+    .replace(',', '')
+    .replace(/[ ]{2,}/g, ' ')
```
eshaham: @esakal maybe add a comment for dummies explaining what this line does. I still don't understand :-)

src/scrapers/leumi-card.js
```diff
@@ -151,10 +150,62 @@ async function getAccountNumber(page, cardIndex) {
...
+  let currency = SHEKEL_CURRENCY;
+  let amount = null;
+  const parts = formattedAmount.split(' ');
+
+  amount = parseFloat(parts[0]);
+  if (parts.length === 2) {
+    currency = fromCurrencySymbolToValue(parts[1]);
+
+    if (currency === null) {
+      throw new Error(`cannot parse amount, failed to detect currency for '${amountStr}'`);
```
eshaham: @esakal we will probably get a lot of txns in currencies other than ILS or USD.
It feels a bit aggressive to throw an error for such an expected input

esakal: well I agree it is n aggressive approach. We actually have a softer alternative. if the scraper fails to detect the currency it can set an empty value. That way if someone is interested in the currency symbol and is using different currency will open an issue and others will not be affected.

This is a reasonable choice, but it is  based no the assumption/decision that failing to extract the currency is acceptable and should fallback to empty value.

any thoughts?

esakal: Hi @eshaham, 
What do you think?

eshaham: @esakal yes, that sounds like a good solution.

esakal:
@eshaham I modified the code to ignore unknown currencies as seen here (to test it I just replaced the actual currency with a junk value)

image

eshaham:
@esakal great. see my [other comment](https://github.com/eshaham/israeli-bank-scrapers/pull/157#discussion_r216159088)

PR #159, opened 2018-10-05
Bank leumi foreign account

esakal:
# Proposed changes
- support Bank Leumi foreign account scraping for USD currency.
- easy extend support for other currencies  in Bank Leumi by updating a single switch...case

This PR replaces #156 as it doesn't relate directly to charged currency. It adds support for a new type of account.

## Extended infrastructure
- utility function to execute page evaluation of puppeteer .
- utility function to execute navigation to page.

### DIFF
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -24,7 +24,7 @@ function getPossibleLoginResults() {
   const urls = {};
   urls[LOGIN_RESULT.SUCCESS] = [/ebanking\/SO\/SPA.aspx/];
   urls[LOGIN_RESULT.INVALID_PASSWORD] = [/InternalSite\/CustomUpdate\/leumi\/LoginPage.ASP/];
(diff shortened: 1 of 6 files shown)


src/scrapers/leumi.js
```diff
@@ -168,22 +187,120 @@ async function fetchTransactionsForAccount(page, startDate) {
...
+
+  return result;
+}
+
 async function fetchTransactions(page, startDate) {
-  // TODO need to extend to support multiple accounts and foreign accounts
-  return [await fetchTransactionsForAccount(page, startDate)];
+  return [
+    await fetchTransactionsForLocalAccount(page, startDate),
+    ...await fetchTransactionsForForeignAccounts(page, startDate),
```
eshaham: @esakal mind extracting the `await` calls into variables and then do `[localTxns, ...foreignTxs]`? Would be cleaner.

PR #161, opened 2018-10-11
bank-leumi: fix invalid url used to access accounts

esakal:
`leumi.js` was navigation to accounts page is broken due to changes in their site. Previously they accepted `//` as part of the url and now they don't.

Trying to run the scrape fails with message `Error: failed to find element matching selector "select#ddlTransactionPeriod"`

@eshaham I think you should consider deploying a new version as it is currently not possible to scrape bank leumi without this fix.

### DIFF
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -63,7 +63,7 @@ class BaseScraperWithBrowser extends BaseScraper {
     if (this.options.verbose) {
       env = Object.assign({ DEBUG: '*' }, process.env);
     }
-    
+
     if (typeof this.options.browser !== 'undefined' && this.options.browser !== null) {
       this.browser = this.options.browser;
     } else {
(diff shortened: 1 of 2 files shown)


src/scrapers/base-scraper-with-browser.js
```diff
@@ -63,7 +63,7 @@ class BaseScraperWithBrowser extends BaseScraper {
     if (this.options.verbose) {
       env = Object.assign({ DEBUG: '*' }, process.env);
     }
-    
+
```
esakal: I needed to remove a space here which caused jslint pre-commit to fail.

src/scrapers/leumi.js
```diff
@@ -10,7 +10,7 @@ import {
 import { waitForNavigation } from '../helpers/navigation';
 import { SHEKEL_CURRENCY, NORMAL_TXN_TYPE, TRANSACTION_STATUS } from '../constants';
 
-const BASE_URL = 'https://hb2.bankleumi.co.il/';
+const BASE_URL = 'https://hb2.bankleumi.co.il';
```
esakal: this will result with a valid url without duplicate `/` symbol after the base url.

`https://hb2.bankleumi.co.il/ebanking/Accounts/ExtendedActivity.aspx?WidgetPar=1#/`

ifeins:
Thanks @esakal, I was having issues with the leumi scraper and not enough time to look into it 👍 

@eshaham Can you release a new version and integrate it to `israeli-ynab-updater`?

eshaham:
@ifeins done

Issue #167, opened 2018-11-05
Leumi-card scraper ends with error


mellowcandle:
Leumi Card: START_SCRAPING
Leumi Card: INITIALIZING
Leumi Card: LOGGING_IN
Leumi Card: LOGIN_SUCCESS
Leumi Card: TERMINATING
Leumi Card: END_SCRAPING
success: false
error type: generic
error: Unknown transaction type קרדיט

mellowcandle:
Merged, closing issue

PR #168, opened 2018-11-05
leumiCard: Add installment type: קרדיט

mellowcandle:
Fix bug in leumiCard causing scraper to fail.
#167


src/scrapers/leumi-card.js
```diff
@@ -66,6 +67,7 @@ function getTransactionType(txnTypeStr) {
     case MONTHLY_CHARGE_PLUS_INTEREST_TYPE_NAME:
       return NORMAL_TXN_TYPE;
     case INSTALLMENTS_TYPE_NAME:
+    case CREDIT_TYPE_NAME:
```
esakal: Hi, Thanks for your contribution. I just want to confirm that קרדיט should be treated as installments (תשלומים)? isn't it a one time payment? If it is then it should result with `NORMAL_TXN_TYPE`. wdyt?

mellowcandle:
Credit is an installment type with additional fee.

PR #174, opened 2019-02-18
leumi: detect password issue blocker and return proper message

esakal:
# Motivation
Leumi site warn after login if password is going to be expired. in such a scenario we currently wait until timeout is raised.

A more comprehensive solution will be to map such a scenario to 'password expired' state but it will require changing `base-scraper-with-browser` to let the scraper interfere with the `handleLoginResult` process. 
A much simpler solution will be to map it to 'password invalid' which doesn't require changes in the base scraper. @eshaham let me know if you agree with this simple adjustment.
 
image

image

### DIFF
--- a/src/scrapers/leumi.js
+++ b/src/scrapers/leumi.js
@@ -191,6 +191,7 @@ async function waitForPostLogin(page) {
   return Promise.race([
     waitUntilElementFound(page, 'div.leumi-container', true),
     waitUntilElementFound(page, '#loginErrMsg', true),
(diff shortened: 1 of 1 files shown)


esakal:
@eshaham unfortunately I will need a better wait selector to be used here but I already changed my password. Will create a new PR once it will happen again.

eshaham:
@esakal 👍 
I don't mind either of the 2 proposed solutions, would probably go with the simpler solution.

PR #176, opened 2019-02-27
Fix leumi card stuck in popup before login

brafdlog:
This PR fixes a bug that Leumi card scraper currently fails to login because they added a popup that you need to close before logging in.

Changes:
- Added support for a generic pre login hook in base-scraper-with-browser
- Made the leumi card scraper run a pre login hook that closes the popup that prevents login.

### DIFF
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -119,6 +119,9 @@ class BaseScraperWithBrowser extends BaseScraper {
     }
 
     await this.fillInputs(loginOptions.fields);
+    if (loginOptions.preAction) {
+      await loginOptions.preAction();
+    }
     await clickButton(this.page, loginOptions.submitButtonSelector);
     this.emitProgress(SCRAPE_PROGRESS_TYPES.LOGGING_IN);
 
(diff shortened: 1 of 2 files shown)


brafdlog:
@eshaham Can you take a look?

src/scrapers/base-scraper-with-browser.js
```diff
@@ -119,6 +119,9 @@ class BaseScraperWithBrowser extends BaseScraper {
     }
 
     await this.fillInputs(loginOptions.fields);
+    if (loginOptions.preAction) {
```
esakal: @brafdlog as this is a pre action that might serve other scenarios, there might be situations where the `loginOptions.fields` will not exist on the dom until competing the pre action phase. consider moving this block above line 121 (above `await this.fillInputs(loginOptions.fields);`)

brafdlog: Good point. Fixed

esakal:
@eshaham let me know if I can merge it.

brafdlog:
In the last days leumicard made additional changes to their website which broke the scraper again.
Will try to find a time soon to create another PR to fix the scraper to work with these new changes.

esakal:
@brafdlog please see #180.

Issue #171, opened 2019-01-22
Getting extra input from the user


Urigo:
I'm working on the business version of Hapoalim.

I got it all running locally including getting the foreign and credit card transactions

The thing is, that in a business account once you have a successful login they send you SMS with a code.
I've added a command prompt locally to get it working easily for me, but I wonder what would you think would be the best way to add something like that in the current way the code and the login flow is set up?

thanks

esakal:
Hi @Urigo,
I saw your question earlier today. While playing with our codebase I ended up implementing this feature as it felt natural extension to the login process.

@eshaham can you please review my solution in pr #172, I think we achieve an important support with efficient code extension. It is not ready yet until @Urigo will create a PR to branch `sms-verification` with the relevant Hapoalim selectors/url but beside that it seems ready.

@Urigo as said above, please pull `sms-verification`. I added `TODO Urigo: ...` comments in `hapoalim.js` file as placeholders. Once it works just create a PR with your additions. If you are using the Playground support to test your scraper, I already extended the playground to support it, otherwise assign `smsVerificationHandler` with a relevant handler

Any insights?

Urigo:
hi again and sorry for the very late response.

@esakal thank you so much for the PR, it was very clear.

The reason I haven't responded was that I was focusing on finishing the project that uses the new scraper so I would make sure I run it every day.
Now it is the case and I am very happy with it :)

Now I want to contribute it back to the codebase.

The main question I have is that (and probably it should go into a separate issue) - how import and helpful is the current class structure for you?

My very personal opinion while working on the scraper, even though it was probably the most similar scraper out there to another scraper (Poalim), have the class setup just made it harder.
At some point I just converted all functions to be stateless and just created by own flow, some of it similar to the current Poalim and some not.

The functional helper directory was very helpful and I kind of expended it with code from the current Poalim scraper and just created my own flow instead of confirming to the flow dictated but the base class.

I understand the idea of exposing the same API from all scrapers in order to have a single runner with similar status and steps updates, but what is that runner?

I think each developer would probably create their own runner in case they will create a general application and that runner will expect statuses, steps and errors that are specific to that parent application.

It is probably not my place to ask for it but have you thought about changing the structure into a lot of helper functions, some specific to a bank, some general and some just wrapping a bunch of other functions for a mini-flow?

Urigo:
[...]

PR #186, opened 2019-03-24
Provide lego blocks for powerful flow creation

esakal:
@eshaham @Urigo Hi,
I was thinking about what Uri wrote [here](https://github.com/eshaham/israeli-bank-scrapers/issues/171#issuecomment-471001945), quoting: 
> ... all functions to be stateless and just created by own flow ...

Currently our code base is tight to one specific flow (1. login   2. scrape transactions). We can easily refactor the internal logic to have stateless representation of this flow and provide much powerful infrastructure:
1. The existing public api remains untouched, one can run any scrapers as he is currently doing and get the same api, emitted events and obviously response type.
2. Anyone can create his own flow now that our library will provide lego blocks that are easy to reuse.

Some thoughts about my suggestion:
1. I only refactored Bank Leumi. It was quite straightforward and if you wish I will be happy to demonstrate the same practice on Leumi card.
2. We will be able to provide many new use-cases which anyone can pick and choose from. Some examples can be balance scraping, checks scraping, sms verification and foreign accounts.
3. As any new use-case will be provided independently, we no longer need to have 100% support in all scrapers before they can reach our 'master' branch.  
3. Looking forward, if we want to allow integration of this library in electron or other hosted environments, we should consider using [puppeteer-core](https://github.com/GoogleChrome/puppeteer#puppeteer-core).
4. Another missing essential feature with our existing scrapers is ability to cancel scraping invocation. Although cancellable promises will not be added to future ECMA releases, It  should be easily added by using second promise provided in the options. 

My suggestion is to pull this PR and view the files from IDE instead of using the Github PR changes view. Using the IDE better demonstrate the motivation behind this PR. 

wdyt?
Eran.


esakal:
@eshaham @urigo wdyt?

esakal:
closing this PR, will create a new on later

PR #187, opened 2019-05-18
Fixed leumi card truncated transaction descriptions

orzarchi:
When scraping leumi card/MAX forex transaction, the description gets truncated.
This change brings the full description from a hidden box that is usually displayed when the transaction is clicked.

Screen Shot 2019-05-18 at 21 03 06

### DIFF
--- a/src/scrapers/leumi-card.js
+++ b/src/scrapers/leumi-card.js
@@ -164,9 +164,8 @@ async function getTransactionsForSection(page, cardIndex, sectionIndex) {
   const txnsRows = await cardSections[sectionIndex].$$('.jobs_regular');
   const expandedBusinessesNamesHeaders = await cardSections[sectionIndex].$$('.openedJob .bisName');
   let expandedBusinessesNames = await Promise.all(expandedBusinessesNamesHeaders.map(
-    header => page.evaluate(y => y.innerText, header),
-  ),
-  expandedBusinessesNamesHeaders);
+    header => page.evaluate(x => x.innerText, header),
+  ), expandedBusinessesNamesHeaders);
 
   // Leumicard keeps hidden open transactions without any content, filter them out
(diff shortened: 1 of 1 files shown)


src/scrapers/leumi-card.js
```diff
@@ -162,6 +162,15 @@ async function getAccountNumber(page, cardIndex) {
 async function getTransactionsForSection(page, cardIndex, sectionIndex) {
   const cardSections = await getCardSections(page, cardIndex);
   const txnsRows = await cardSections[sectionIndex].$$('.jobs_regular');
+  const expandedBusinessesNamesHeaders = await cardSections[sectionIndex].$$('.openedJob .bisName');
+  let expandedBusinessesNames = await Promise.all(expandedBusinessesNamesHeaders.map(
+    header => page.evaluate(y => y.innerText, header),
+  ),
+  expandedBusinessesNamesHeaders);
```
eshaham: @orzarchi can you please remove the line break and place this in the previous line?

src/scrapers/leumi-card.js
```diff
@@ -186,9 +195,8 @@ async function getTransactionsForSection(page, cardIndex, sectionIndex) {
       return td.innerText;
     }, txnColumns[6]);
 
-    const description = await page.evaluate((td) => {
-      return td.innerText;
-    }, txnColumns[3]);
+    // take the description from the hidden expanded info-card, as the one in the table is truncated
```
eshaham: @orzarchi no need for the comment, this is self explanatory

src/scrapers/leumi-card.js
orzarchi: I was struggling with the linting rules for a few minutes.
May I suggest implementing prettier and eslint autofix as a pre-commit hook?
I can add it to my PR or create a new one.

PR #199, opened 2019-05-24
Error handling

eshaham:
Some refactoring around error handling, plus fixing the annoying timeout error which wasn't really handled once and for all

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -27,7 +27,7 @@ export const LOGIN_RESULT = {
   UNKNOWN_ERROR: 'UNKNOWN_ERROR',
 };
 
-export const NAVIGATION_ERRORS = {
+export const ERRORS = {
   TIMEOUT: 'TIMEOUT',
   GENERIC: 'GENERIC',
 };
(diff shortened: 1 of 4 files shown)


esakal:
The code changes looks relevant. I remember that we have an issue with an error not being propagated correctly but I failed to reproduce it now (don't remember the exact scenario). 

I tested both valid scenario and invalid scenario (providing false credentials to Bank Leumi) and both worked as expected. Note that we still get `generic` error type when having timeout 
```
Bank Leumi: START_SCRAPING
Bank Leumi: INITIALIZING
Bank Leumi: TERMINATING
Bank Leumi: END_SCRAPING
success: false
error type: generic
error: Navigation Timeout Exceeded: 30000ms exceeded
```

eshaham:
@esakal 😱 
how did you reproduce the timeout?

esakal:
The wifi on the train is poor so I usually get timeouts when scraping. I can debug this PR on Wed if you want. Another options is using https://www.charlesproxy.com/ if I remember it has an option to simulate timeout

esakal:
@eshaham I think you should merge this pr as-is and ignore the edge case I mentioned. It seems to improve the error handling and better describe what happens.

eshaham:
@esakal 👍  I agree

PR #201, opened 2019-05-25
Attempt to Fix Visa Cal scraper

ifeins:
I've applied the changes suggested by @sagiba in https://github.com/eshaham/israeli-bank-scrapers/issues/200 but I'm still unable to get the scraper to work.

It fails with the following output:

```
➜  israeli-bank-scrapers git:(fix_visa_cal_scraper) ✗ npm start

> israeli-bank-scrapers@0.6.7 start /Users/ifeins/Development/israeli-bank-scrapers
> node -r babel-register playground/scrape.js

Visa Cal: START_SCRAPING
Visa Cal: INITIALIZING
Visa Cal: LOGGING_IN
Auth result: "שם המשתמש או הסיסמה שהוזנו שגויים"
Visa Cal: TERMINATING
Visa Cal: END_SCRAPING
success: false
error type: generic
error: body used already for: https://connect.cal-online.co.il/api/authentication/login
```

@sagiba Can you let me know if I've missed anything?

PR #202, opened 2019-05-25
Fix visa cal scraper

sagiba:
More fixes based on @ifeins previous PR #201

### DIFF
--- a/src/scrapers/visa-cal.js
+++ b/src/scrapers/visa-cal.js
@@ -16,7 +16,7 @@ import {
 import { fetchGet, fetchPost } from '../helpers/fetch';
 import { fixInstallments, sortTransactionsByDate, filterOldTransactions } from '../helpers/transactions';
 
-const BASE_URL = 'https://restservices.cal-online.co.il/Cal4U';
+const BASE_URL = 'https://cal4u.cal-online.co.il/Cal4U';
 const DATE_FORMAT = 'DD/MM/YYYY';
 
 const PASSWORD_EXPIRED_MSG = 'תוקף הסיסמא פג';
@@ -34,6 +34,8 @@ const MEMBERSHIP_FEE_TYPE_CODE = '67';
 const SERVICES_REFUND_TYPE_CODE = '71';
 const SERVICES_TYPE_CODE = '72';
 const REFUND_TYPE_CODE_2 = '76';
+
+const HEADER_SITE = { 'X-Site-Id': '8D37DF16-5812-4ACD-BAE7-CD1A5BFA2206' };
 
 function getBankDebitsUrl(accountId) {
   const toDate = moment().add(2, 'months');
@@ -145,19 +147,19 @@ function prepareTransactions(txns, startMoment, combineInstallments) {
 
 async function getBankDebits(authHeader, accountId) {
   const bankDebitsUrl = getBankDebitsUrl(accountId);
-  return fetchGet(bankDebitsUrl, authHeader);
+  return fetchGet(bankDebitsUrl, Object.assign(authHeader, HEADER_SITE));
 }
 
 async function getTransactionsNextPage(authHeader) {
   const hasNextPageUrl = `${BASE_URL}/CalTransNextPage`;
-  return fetchGet(hasNextPageUrl, authHeader);
+  return fetchGet(hasNextPageUrl, Object.assign(authHeader, HEADER_SITE));
 }
 
 async function fetchTxns(authHeader, cardId, debitDates) {
   const txns = [];
   for (const date of debitDates) {
     const fetchTxnUrl = getTransactionsUrl(cardId, date);
-    let txnResponse = await fetchGet(fetchTxnUrl, authHeader);
+    let txnResponse = await fetchGet(fetchTxnUrl, Object.assign(authHeader, HEADER_SITE));
     if (txnResponse.Transactions) {
       txns.push(...txnResponse.Transactions);
     }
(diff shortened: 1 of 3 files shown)


src/scrapers/visa-cal.js
```diff
@@ -250,14 +253,9 @@ class VisaCalScraper extends BaseScraper {
...
-    if (_.get(authResponse, 'Response.Status.Succeeded')) {
-      this.authHeader = `CalAuthScheme ${authResponse.AuthenticationToken}`;
-      this.emitProgress(SCRAPE_PROGRESS_TYPES.LOGIN_SUCCESS);
-      return { success: true };
-    }
-
-    throw new Error('unknown error during login');
+    this.authHeader = `CALAuthScheme ${authResponse.token}`;
+    this.emitProgress(SCRAPE_PROGRESS_TYPES.LOGIN_SUCCESS);
+    return { success: true };
```
ifeins: Isn't there a way to know if the authentication failed?
Because here it seems that it will always return success regardless of whether the authentication succeeded.

sagiba: The two blocks above check if Response.Status.Message exist in the payload, which appears to be the indication for a failure. Looking at it again now, I would add another generic error after line 254, that aborts if the "token" parameter does not exist and we couldn't match any of the more specific error messages.

esakal: It will be a good guard making sure the 'token' exists

src/scrapers/visa-cal.js
```diff
@@ -145,19 +147,19 @@ function prepareTransactions(txns, startMoment, combineInstallments) {
 
 async function getBankDebits(authHeader, accountId) {
   const bankDebitsUrl = getBankDebitsUrl(accountId);
-  return fetchGet(bankDebitsUrl, authHeader);
+  return fetchGet(bankDebitsUrl, Object.assign(authHeader, HEADER_SITE));
```
esakal: Can you please revert all 4 occurrences that re-assign `authHeader`: 
1. revert this change so you will pass `authHeader` as is.
2. change line 274 in method `fetchData`  from 
```
    const authHeader = { Authorization: this.authHeader };	
```
 to 
```
    const authHeader = Object.assign({ Authorization: this.authHeader }, HEADER_SITE);
```

sagiba:
I've updated the PR with all the fixes discussed above.

ifeins:
Looks good to me, thanks for submitting the PR 👍 
@eshaham Can you or one of the admins review this PR?

ifeins:
I've locally pulled the branch but when I try to run it in `israeli-bank-scrapers` playground I get the following error:

```
➜  israeli-bank-scrapers git:(fix_visa_cal_scraper) npm start

> israeli-bank-scrapers@0.6.7 start /Users/ifeins/Development/israeli-bank-scrapers
> node -r babel-register playground/scrape.js

Visa Cal: START_SCRAPING
Visa Cal: INITIALIZING
Visa Cal: LOGGING_IN
Visa Cal: TERMINATING
Visa Cal: END_SCRAPING
success: false
error type: UNKNOWN_ERROR
error: undefined
```

Does the `israeli-bank-scrapers` playground works for you with the changes in this branch?

sagiba:
Works for me. Not sure what error: undefined means. Can you try adding some debug messages and see if you can figure out the what causes this?

sagiba:
Please pull and try now, I added more verbose error logging after login failures.

esakal:
@sagiba it seems that when error happen they return a simple string. so you should do something like:
```
const authResponse = await fetchPost(authUrl, authRequest, HEADER_SITE);

    if (authResponse === PASSWORD_EXPIRED_MSG) {
      return {
        success: false,
        errorType: LOGIN_RESULT.CHANGE_PASSWORD,
      };
    }

    if (authResponse === INVALID_CREDENTIALS) {
      return {
        success: false,
        errorType: LOGIN_RESULT.INVALID_PASSWORD,
      };
    }

    if (!authResponse || !authResponse.token) {
      return {
        success: false,
        errorType: LOGIN_RESULT.UNKNOWN_ERROR,
      };
    }
    this.authHeader = `CALAuthScheme ${authResponse.token}`;
    this.emitProgress(SCRAPE_PROGRESS_TYPES.LOGIN_SUCCESS);
    return { success: true };
```

also they changed the error message for invalid password so to make it works you should change line 23 to:
```
const INVALID_CREDENTIALS = 'שם המשתמש או הסיסמה שהוזנו שגויים';
```
It is easy to test it using the playground, just provide invalid user/password

sagiba:
It's a little more complicated than that: because the response format is now inconsistent (either text or json), we cannot use fetchPost as-is because it tries to parse the JSON payload internally, and throws an exception if it failed.

We need to make it optionally return the text or the raw result object and parse the JSON in the login method itself.

esakal:
I was also surprised that it works. I don't get get internal error, if it 
happens it doesn't stop the flow. Also, try to parse an invalid response to 
text and you will get wrapped string " '...' ". Not sure why they chose to 
implement it like that but again it works on my machine. Do you experience 
a different behaviour?

sagiba:
Ah, I see what happens. They return the error message when it's quoted so it is in fact a valid JSON, and manages to parse correctly. So your code above is correct. I will push in a minute.

Unfortunately, my account is now blocked :-( and we now know one of their other error messages:
"הכניסה למנוי נחסמה עקב ריבוי נסיונות כושלים. לשחרור המנוי באפשרותך לחדש סיסמה על ידי בחירת שכחתי שם משתמש סיסמה"

ifeins:
@sagiba I've pulled the latest changes and re-ran it, I got the following error:

```
➜  israeli-bank-scrapers git:(fix_visa_cal_scraper) npm start

> israeli-bank-scrapers@0.6.7 start /Users/ifeins/Development/israeli-bank-scrapers
> node -r babel-register playground/scrape.js

Visa Cal: START_SCRAPING
Visa Cal: INITIALIZING
Visa Cal: LOGGING_IN
Visa Cal: TERMINATING
Visa Cal: END_SCRAPING
success: false
error type: UNKNOWN_ERROR
error: No token found in authResponse: "שם המשתמש או הסיסמה שהוזנו שגויים"
```

I changed my password in Visa Cal site and set up the credentials again and was surprised to get the same error.

sagiba:
@ifeins Try the Cal4U mobile app and see if you can login there. It uses the same API.

esakal:
@sagiba thanks for your effort!
@eshaham can you deploy a new version?

Issue #80, opened 2018-02-20
Add a charged currency property to transactions


eshaham:
In some cases when users have a foreign currency account in their banks, transactions in that foreign currency get charged directly against the foreign currency bank account, and conversion to `ILS` is not needed.

Currently this module always assumes `chargedAmount` is returned in `ILS`.
We should introduce a new `chargedCurrency` property to transactions that will hold the actual currency used.

asfktz:
I would like to handle that for Isracard and Bank Hapoalim.

One challenge I have it that I'm not sure I've seen a case where a transaction in foreign currency is converted to ILS.

In my Isracard account, I see that some of my transactions are converted to ILS, but the "Original Currency" column is also ILS:

(Isracard)
![image](https://user-images.githubusercontent.com/199747/36499539-3dcf9ae4-174a-11e8-9155-9fffcaa335cd.png)

Note: My credit card is set to transfer money from my `מט"ח` account for foreign currency

Do you have an example of such case?

esakal:
Hi @asfktz, following eshaham/israeli-ynab-updater#22 you mentioned that the transaction contain the following information:
```
{
  "currencyId": "USD",
  "currentPaymentCurrency": "USD",
}
```

can't we use those fields to figure original and charged currency?

Regarding the banks I think we can scrape also the foreign accounts, I can add the Leumi account if needed

esakal:
@asfktz I would like to promote the support for charged currency. Please take a look at #155, I added new field to `leumi-card` named `chargedCurrency` in [charged-currency branch](https://github.com/eshaham/israeli-bank-scrapers/tree/charged-currency). If you can add their support for `Isracard` it would be great.

asfktz:
Better late than never.
I'm working on the `chargedCurrency` field for `Isracard`.


First here are examples of the two transaction types

## Transaction Examples

### 1. A transaction that is charged in foreign currency:
(It charges my foreign currency account directly)

![image](https://user-images.githubusercontent.com/199747/59962957-275fd100-94f5-11e9-8fa5-1796ca33784a.png)

```
{
    "specificDate": null,
    "cardIndex": "0",
    "dealsInbound": null,
    "supplierId": null,
    "supplierName": null,
    "dealSumType": null,
    "paymentSumSign": null,
    "purchaseDate": null,
    "fullPurchaseDate": null,
    "moreInfo": null,
    "horaatKeva": null,
    "voucherNumber": "0069338",
[...]

Issue #155, opened 2018-09-01
scrape charged currency for credit cards


esakal:
We are missing charged currency when scraping credit cards. This information is needed as the user can choose whether to link his credit card to local bank account or foreign bank account. Without charged currency the value of charged amount means nothing. 

I implemented that support for `leumi card scraper` in branch [charged-currency](https://github.com/eshaham/israeli-bank-scrapers/tree/charged-currency). the following shows the scraping of my card from with support to all 3 use cases (`ILS->ILS`, `USD->ILS`, `USD->USD`):
![image](https://user-images.githubusercontent.com/4751797/44949712-d258ae80-ae3f-11e8-88a3-e29d0ca00eb0.png)

@eshaham,  If you find it relevant we can use that branch to add support for other cards. wdyt?

esakal:
Yes it does :) I missed that one. So great I will continue and add support for `leumi` foreign account and hopefully @asfktz will add support for Poalim and Isracard

asfktz:
Nice! I'm happy to see that.
I have so much love for this project (:

Yeah, I'll be able to add support for Poalim and Isracard.
I'm not available for open source so much lately, so it might take some time.

esakal:
closing this issue as it is a duplicate of #80

PR #214, opened 2019-07-07
chargedCurrency for Isracard

asfktz:
related to #80

### DIFF
--- a/src/scrapers/base-isracard-amex.js
+++ b/src/scrapers/base-isracard-amex.js
@@ -110,6 +110,7 @@ function convertTransactions(txns, processedDate) {
       originalAmount: isOutbound ? -txn.dealSumOutbound : -txn.dealSum,
       originalCurrency: convertCurrency(txn.currencyId),
       chargedAmount: isOutbound ? -txn.paymentSumOutbound : -txn.paymentSum,
+      chargedCurrency: convertCurrency(txn.currencyId),
       description: isOutbound ? txn.fullSupplierNameOutbound : txn.fullSupplierNameHeb,
       memo: txn.moreInfo,
       installments: getInstallmentsInfo(txn),


esakal:
@asfktz Hi,
I don't think any scraper currently support `chargedCurrency` field. I don't see any mentioning about it in [CONTRIBUTING.md](https://github.com/eshaham/israeli-bank-scrapers/blob/master/CONTRIBUTING.md) file. Is this a new suggestion? 

Since it currently returns the same value as the original currency and other scrapers don't handle such a field, I'm not sure if we should add it for now. We can obviously consider adding such a field officially if most of the scrapers exposes that value.

@eshaham @asfktz  wdyt?

asfktz:
Hi @esakal 👋

It's related to this issue https://github.com/eshaham/israeli-bank-scrapers/issues/80

eshaham:
@asfktz @esakal looking back at #80, I now remember that **a**, we created a dedicated branch `charged-currency`, although it only has documentation, and **b**, it's not really clear (even in Isracard's case) that the transaction was charged in the original currency.
While there's definitely a need, I'm not sure we can really support it across the scrapers.
Question is, do you guys think we should try to do our best to support it anyway?

esakal:
@eshaham I agree that it is not that obvious feature to implement across all scrapers, mostly because the contributors seems to charge transactions in local currency (ILS) so it will be difficult to integrate it to all our scrapers and maintain it. 

Just to point out that we already have pending support for Leumi card as well here #155. I did it back in the days when I charged my cards with foreign currency and needed it to reconsolidate my accounts.

I do think that we need to have a better adaption strategy that will allow gradually feature adoption by marking some features as optional. That way any contributor will be able to extend the library and have it available in production in no time instead of ignoring them due to lack of support in other scrapers. The `readme.md` file should be extended to include a feature-scraper support matrix, encouraging contributors to extend the library where possible, something as followed:

| Feature | Bank Hapoalim | Bank Leumi |  Bank Discount | Bank Otsar Hahayal | Visa Cal |  Leumi-Card | Isracard | Amex |
| --- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Local accounts | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
| Installments | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
| Pending Status | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
| Charged currency | **No** | **No** | **No** | **No** | **No** | **No** | Yes | Yes | **No** |

This is only an example, obviously can make it more attractive by replacing the `yes` `no` text with visual icon. The point is that we will be more flexible in adding new features by saying that some features are must have in all scrapers while others are optional.

wdyt?

asfktz:
> I do think that we need to have a better adaption strategy that will allow gradually feature adoption by marking some features as optional.

@esakal I like that! 👍 


> looking back at #80, I now remember that a, we created a dedicated branch charged-currency, although it only has documentation, and b, it's not really clear (even in Isracard's case) that the transaction was charged in the original currency.
While there's definitely a need, I'm not sure we can really support it across the scrapers.
Question is, do you guys think we should try to do our best to support it anyway?

It's a valuable property to have in my option.
I'm a bit concern that the charged currency is actually not available for us in Isracard and we provide from data (originalCurrency instead) https://github.com/eshaham/israeli-bank-scrapers/issues/80#issuecomment-509014942, see why [here](https://github.com/eshaham/israeli-bank-scrapers/issues/80#issuecomment-504664505).

Maybe it's better to wait with the Isracard implementation for now?

In the meantime time, can we implement the currency convetion for Isreali-ynab-updater 
using original currency instead until we figure out what to do?

The implementation is ready, I only need to fix conflicts with master.
https://github.com/eshaham/israeli-ynab-updater/pull/26

Issue #231, opened 2019-07-28
Leumi scraper throws an error when there are no transactions during the given date range


brafdlog:
To find transactions the leumi scraper looks for the following selector: `table#WorkSpaceBox table#ctlActivityTable`. 

When there are no transaction in a specific date range, the table that this selector points to does not exist and instead the following message is shown:

<img width="666" alt="Screenshot 2019-07-28 17 07 28" src="https://user-images.githubusercontent.com/7272927/62007986-f3637580-b15c-11e9-8ac8-35f3ddf85980.png">

Since the selector does not exist, the scraper times out and throws an error.

brafdlog:
Created a pr #232

brafdlog:
@esakal Can you take a look at the pr?

esakal:
fixed and merged  #232

PR #232, opened 2019-07-28
Fix issue #231 - leumi scraper throws error when no transactions in date range

brafdlog:


brafdlog:
I have actually seen this a while ago but didn't get around to making a PR until now.

Your suggestion is really good. I actually thought my solution was not optimal there but didn't have a better solution in mind.

Will update the PR accordingly.

brafdlog:
@esakal Pushed a fix according to your comment

esakal:
@brafdlog it looks great. 👍 

@eshaham please deploy a new version when you can.

PR #234, opened 2019-08-09
Leumi Card failed because next month is null

baruchiro:
When I scrape the Leumi Card, the next month request returns with `data.result` is `null`.

The test failed before the repair and is now passing.

PR #236, opened 2019-08-11
Fix: Leumi card throws when there are no future transactions

### LINKED ISSUE #237: Leumi card throws when there are no future transactions
Leumi card throws when there are no future transactions, at least I guess this is the case.
What I can tell for sure is that it falls in `fetchTransactionsForMonth` when `data.result` is `null`.
In my case, it happens when `monthMoment` is next month.

https://github.com/eshaham/israeli-bank-scrapers/blob/59b5056e2e66ff501c3861e0b19dc2c2b76949af/src/scrapers/leumi-card.js#L114-L124

PR https://github.com/eshaham/israeli-bank-scrapers/pull/236 fixes it by adding a null check.

asfktz:
Fixes #237

### DIFF
--- a/src/scrapers/leumi-card.js
+++ b/src/scrapers/leumi-card.js
@@ -1,5 +1,6 @@
 import buildUrl from 'build-url';
 import moment from 'moment';
+import _ from 'lodash';
 import { fetchGetWithinPage } from '../helpers/fetch';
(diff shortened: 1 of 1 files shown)


esakal:
@asfktz Hi, This same issue was fixed by @baruchiro in #234 and should be deploy soon by @eshaham into NPM. Please let me know if it doesn't solve your issue.

asfktz:
Oh, nice!
Ok, I'll close this one.

PR #245, opened 2019-08-16
Check if the url is updated in navigateTo

### LINKED ISSUE #244: Hash navigation (Angular router) return null and fail
https://github.com/eshaham/israeli-bank-scrapers/blob/59b5056e2e66ff501c3861e0b19dc2c2b76949af/src/scrapers/base-scraper-with-browser.js#L94-L100

In this code, when we trying to navigate to a page that differs only after the hash, the request occurs in Ajax and there are no page loading events, so Puppeteer doesn't know that it was a request.

So the `response` will be `null` and the function will throw.

baruchiro:
closes #244 

Tested with `Leumi` and `Leumi Card`.

### DIFF
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -94,7 +94,10 @@ class BaseScraperWithBrowser extends BaseScraper {
   async navigateTo(url, page) {
     const pageToUse = page || this.page;
     const response = await pageToUse.goto(url);
-    if (!response || response.status() !== OK_STATUS) {
+
+    const responseSuccess = response && response.status() === OK_STATUS;
+    const navigated = responseSuccess || pageToUse.url() === url;
+    if (!navigated) {
       throw new Error(`Error while trying to navigate to url ${url}`);
     }
   }
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -95,9 +95,7 @@ class BaseScraperWithBrowser extends BaseScraper {
     const pageToUse = page || this.page;
     const response = await pageToUse.goto(url);
 
-    const responseSuccess = response && response.status() === OK_STATUS;
-    const navigated = responseSuccess || pageToUse.url() === url;
-    if (!navigated) {
+    if (response !== null && response.status() !== OK_STATUS) {
       throw new Error(`Error while trying to navigate to url ${url}`);
     }
   }
--- a/src/scrapers/base-scraper-with-browser.js
+++ b/src/scrapers/base-scraper-with-browser.js
@@ -94,8 +94,9 @@ class BaseScraperWithBrowser extends BaseScraper {
   async navigateTo(url, page) {
     const pageToUse = page || this.page;
     const response = await pageToUse.goto(url);
-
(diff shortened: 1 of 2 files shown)


src/scrapers/base-scraper-with-browser.js
```diff
@@ -94,7 +94,10 @@ class BaseScraperWithBrowser extends BaseScraper {
   async navigateTo(url, page) {
     const pageToUse = page || this.page;
     const response = await pageToUse.goto(url);
-    if (!response || response.status() !== OK_STATUS) {
+
+    const responseSuccess = response && response.status() === OK_STATUS;
+    const navigated = responseSuccess || pageToUse.url() === url;
```
esakal: hi @baruchiro, reading [puppeteer goto documentation](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options]

>  NOTE page.goto either throws an error or returns a main resource response. The only exceptions are navigation to about:blank or navigation to the same URL with a different hash, which would succeed and return null.

I'm not sure this is the right condition. it might just work but from the documentation above is seems that the condition should be `response === null || response.status() === OK_STATUS`.  wdyt?

baruchiro: You're right.
But now I think the only case where we'll have to throw an error is `response.status !== OK_STATUS`
Because in all other cases (`response === null || response.status() === OK_STATUS`, the error will be thrown by `goto`, or we got the expected result.

src/scrapers/base-scraper-with-browser.js
```diff
@@ -94,7 +94,8 @@ class BaseScraperWithBrowser extends BaseScraper {
   async navigateTo(url, page) {
     const pageToUse = page || this.page;
     const response = await pageToUse.goto(url);
-    if (!response || response.status() !== OK_STATUS) {
+
+    if (response !== null && response.status() !== OK_STATUS) {
```
esakal: this should work since in puppeteer documentation they mention that when changing hash a null result is returned. but if they will return `undefined` here for any reason we will have unhandled exception.

@eshaham please review previous commit in this pr and the comment I wrote and advice how to proceed.

baruchiro:
In your first comment you highlighted the point of the documentation, and
you were right, `null` meaning there was proper navigation, even if it
wasn't intuitive, because otherwise an error would have been thrown.

So we works with the puppeteer documentation. Now, You ask what will
happen  if `goto` will return `undefined`? This is not expected and will
never happen, so if this happens and an exception is thrown because of our
code, it is right, that our code should not expect it and should throw an
exception.

מקווה שהבנת.

esakal:
I got it :) and you are right, I conflicted my self in those two comments. It is just because I don't think our code should always make sure an object is not undefined before calling its' property. 
how about enjoying both worlds:
```
// note: response will be null when navigating to same url while changing the hash part. the condition below will always accept null as valid result. 
if (response !== null && (response === undefined || response.status() !== OK_STATUS) )
{
}
```
but I agree this is not that big deal, just for safety and clarity of the code

baruchiro:
I don't like the logical expression that includes parentheses, but I see no other way.

baruchiro:
@esakal @eshaham Who merges it?

Issue #253, opened 2019-08-21
Replace button.click with page.$eval


baruchiro:
https://github.com/eshaham/israeli-bank-scrapers/blob/59b5056e2e66ff501c3861e0b19dc2c2b76949af/src/helpers/elements-interactions.js#L13-L16

Using [this site](https://try-puppeteer.appspot.com/), I try two options for clicking a link, and only one works.

**Doesn't work**:
```node
const browser = await puppeteer.launch();

const page = await browser.newPage();
await page.goto('https://example.com');

const button = await page.$('a');
await button.click();

await page.screenshot({path: 'screenshot.png'});

await browser.close();
```

**Works**:
```node
const browser = await puppeteer.launch();

const page = await browser.newPage();
await page.goto('https://example.com');

await page.$eval('a', el => el.click());

await page.screenshot({path: 'screenshot.png'});

await browser.close();
```

Related to #248

baruchiro:
I don't know if to open a new bug, because I don't know yet if we want to support the `iframe`.

The solution to this bug (in #254 ) still doesn't solve the `iframe` problem.
Try using Puppeteer to click on one of the links to change the image of the ball- http://www.littlewebhut.com/articles/html_iframe_example/.

esakal:
@baruchiro the link you provided in previous comment is not relevant as the iframe only show images and the anchors elements are in the top page. But while trying to play with that one I managed to click inside an iframe. Please try and let me know if it helps:
```
const puppeteer = require('puppeteer');


( async function () {
  const browser = await puppeteer.launch({ headless: false});

  const page = await browser.newPage();
  await page.goto('https://www.w3schools.com/TAgs/tryit.asp?filename=tryhtml_button_test');

  const frame = page.frames()[1];
  const button = await frame.$('button');
  await button.click();

  // await page.screenshot({path: 'screenshot.png'});

  // await browser.close();
}());
```
1. copy the above code to a javascript file, ie `iframe.js`
2. run `node iframe.js`
3. you will see an alert that is shown once the button inside the iframe is being clicked.

wdyt?

baruchiro:
Can you replace the `button` with `a[href="https://www.google.com"]`?
The `button` example from you works for me, but with `a` (and `await frame.$('a')`, of course) it does not work.

eshaham:
@esakal @baruchiro where do we stand with this issue?

esakal:
[...]

PR #254, opened 2019-08-21
Use page.$eval

baruchiro:
Fixing #253

### DIFF
--- a/src/helpers/elements-interactions.js
+++ b/src/helpers/elements-interactions.js
@@ -11,8 +11,7 @@ async function fillInput(page, inputSelector, inputValue) {
 }
 
 async function clickButton(page, buttonSelector) {
-  const button = await page.$(buttonSelector);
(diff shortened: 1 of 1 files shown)


src/helpers/elements-interactions.js
```diff
@@ -11,8 +11,7 @@ async function fillInput(page, inputSelector, inputValue) {
 }
 
 async function clickButton(page, buttonSelector) {
-  const button = await page.$(buttonSelector);
-  await button.click();
+  await page.$eval(buttonSelector, el => el.click());
```
eshaham: @baruchiro I wonder if we need to change it for all button clicks
I understand it is required in this use case, I would probably create a new function under `elements-interactions` for clicking an `a` element

baruchiro: Okay, let's just start just in case of `a`.

PR #256, opened 2019-08-30
Mizrahi scraper

### LINKED ISSUE #74: Mizrahi Tefahot scraping brainstorming
Hi,

I've tried to reverse engineer scraping of Mizrahi.

They work with the stupid dot net session thing.

Couldn't find a way to do proper scraping. Anyone here tried to scrape them?

baruchiro:
closes #74


baruchiro:
@eshaham @esakal This PR is ready to merge.

- I have one account in Mizrahi, so I don't know how to support multiple accounts.
- Also, I have only about 2-3 types of transactions, so I don't know what each property in the `json` response means.

After the merge, I will open issues for these two limitations to support whoever wants to fix them.

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+  NORMAL_TXN_TYPE,
+  TRANSACTION_STATUS,
+  ISO_DATE_FORMAT,
+} from '../constants';
+import { BaseScraperWithBrowser, LOGIN_RESULT } from './base-scraper-with-browser';
+import { fetchPostWithinPage } from '../helpers/fetch';
+
+const BASE_URL = 'https://www.mizrahi-tefahot.co.il';
+const LOGIN_URL = `${BASE_URL}/he/bank/Pages/Default.aspx`;
+const AFTER_LOGIN_BASE_URL = 'https://mto.mizrahi-tefahot.co.il/ngOnline/index.html#/main/uis/osh/p428/';
```
esakal: If I remember correctly, Mizrahi allow users to customize the first page post login. which means that the url will end differently and `#/main/uis/osh/p428/` is not guaranteed. 

can you please check the settings page and confirm if this is possible, if so you can always use a regexp and rely only on the start of the link before the hash

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+
+    const response = await fetchPostWithinPage(this.page,
+      TRANSACTIONS_REQUEST_URL, data, headers);
+
+    if (response.header.success === false) {
+      return {
+        success: false,
+        errorType: 'generic',
+        errorMessage:
+          `Error fetching transaction. Response message: "${response.header.messages[0].text}"`,
```
esakal: this is only a suggestion feel free to adapt or ignore. Application developers that will use this library most likely show the error as is to the user. The error returned from the server is usually descriptive and localized (Hebrew). When you add a prefix such `Error fetching....` it actually make the UX worse. 

Again only a suggestion from my experience, consider removing your static prefix and return the error message directly.

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+function CreateDataFromRequest(request, optionsStartDate) {
+  const defaultStartMoment = moment().subtract(1, 'years');
+  const startDate = optionsStartDate || defaultStartMoment.toDate();
+  const startMoment = moment.max(defaultStartMoment, moment(startDate));
+
+  const data = JSON.parse(request.postData());
+
+  data.inToDate = moment().format(DATE_FORMAT);
+  data.inFromDate = startMoment.format(DATE_FORMAT);
+  data.table.maxRow = 9999999999;
```
esakal: is this a valid value that they send as well? if not it might annoy their developers, consider changing to the a bit more reasonable number. If they also send this value feel free to ignore this comment.

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+function createHeadersFromRequest(request) {
+  return {
+    mizrahixsrftoken: request.headers().mizrahixsrftoken,
+    'Content-Type': request.headers()['content-type'],
+  };
+}
+
+function convertTransactions(txns) {
+  return txns.map((row) => {
+    const txnDate = moment(row.MC02PeulaTaaEZ).format(ISO_DATE_FORMAT);
```
esakal: As I cannot run this code, I just want to confirm. I think moment deprecated parsing without providing explicit format as second argument unless Mizrahi returns ISO 8601 format and plan to remove the support of other formats. taken from their documentation: 
> Warning: Browser support for parsing strings is inconsistent. Because there is no specification on which formats should be supported, what works in some browsers will not work in other browsers.
>
> For consistent results parsing anything other than ISO 8601 strings, you should use [String + Format](https://momentjs.com/docs/#/parsing/string-format/).

Please confirm the format and if needed use the [String + Format](https://momentjs.com/docs/#/parsing/string-format/) support

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+
+    // TODO: I don't have enough sample transactions to understand the rest of the data.
+    return {
+      type: NORMAL_TXN_TYPE, // can be either 'normal' or 'installments'
```
esakal: you can remove this comment, installments is relevant only for credit cards

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+      // identifier: int, // only if exists
```
esakal: this is the `asmachta` field (אסמכתא), don't they expose it for each row?

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+      processedDate: txnDate, // ISO date string
+      originalAmount: row.MC02SchumEZ,
+      originalCurrency: SHEKEL_CURRENCY,
+      chargedAmount: row.MC02SchumEZ,
+      description: row.MC02TnuaTeurEZ,
+      // installments: {
+      //   number: int, // the current installment number
+      //   total: int, // the total number of installments
+      // },
+      status: TRANSACTION_STATUS.COMPLETED, // can either be 'completed' or 'pending'
```
esakal: Detecting if transaction is in pending mode is actually important as we had a lot of issues with pending transactions in other scrapers. 

It is quite easy to see pending transactions. Try to access your account during the weekend, usually transactions added during the weekend are separated as they are pending approval.

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+    return {
+      type: NORMAL_TXN_TYPE, // can be either 'normal' or 'installments'
+      // identifier: int, // only if exists
+      date: txnDate, // ISO date string
+      processedDate: txnDate, // ISO date string
+      originalAmount: row.MC02SchumEZ,
+      originalCurrency: SHEKEL_CURRENCY,
+      chargedAmount: row.MC02SchumEZ,
+      description: row.MC02TnuaTeurEZ,
+      // installments: {
```
esakal: you can remove this comment, installments is relevant only for credit cards

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+  return txns.map((row) => {
+    const txnDate = moment(row.MC02PeulaTaaEZ).format(ISO_DATE_FORMAT);
+
+    // TODO: I don't have enough sample transactions to understand the rest of the data.
+    return {
+      type: NORMAL_TXN_TYPE, // can be either 'normal' or 'installments'
+      // identifier: int, // only if exists
+      date: txnDate, // ISO date string
+      processedDate: txnDate, // ISO date string
+      originalAmount: row.MC02SchumEZ,
```
esakal: Please confirm that the resulted value is of type number, if not you will need to manually parse it. This is an example to how we handle amount in leumi:
```
  const amount = parseFloat(amountStrCopy);
```

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,119 @@
...
+      originalCurrency: SHEKEL_CURRENCY,
+      chargedAmount: row.MC02SchumEZ,
```
esakal: same commend as for `originalAmmount`

src/scrapers/mizrahi.js
baruchiro: Yes, I thought it should be unique to the transaction. What the bank gives is a unique identifier for a type of transaction (or something similar).

src/scrapers/mizrahi.js
baruchiro: My solution no longer looks very beautiful.
I need to navigate to the "Osh" (עו"ש) url to sniff the transactions request.
But if I don't `waitForNavigation`, the login not complete.

Maybe should I do this in `fetchData`?

src/scrapers/mizrahi.js
baruchiro: I'm not sure I understand you, but you must know that every time the bank return me an error, it always was `General server error` with no details.
I wanted to add an explanation about that, but I couldn't think of a way to write it.

So the important part is the static message, and the response message is in case it has more information.

src/scrapers/mizrahi.js
baruchiro: They send 50, because this is their page size for paging.
I tried and reached that number. approximately. When it is bigger the request fails.

I hope there is no user with more transactions per year, otherwise we will have to make another request.

src/scrapers/mizrahi.js
esakal: I see, ok, I thought it is like other scrapers where they provide details like `שם המשתמש או הסיסמא שגויה`

src/scrapers/mizrahi.js
esakal: did you see this comment?

src/scrapers/mizrahi.js
baruchiro: Yes, It comes as a `double`.

BTW, If this is important, why is there no verification of this in tests?

esakal:
I see that you are currently working on it. drop a message once you complete your changing.

src/scrapers/mizrahi.js
baruchiro: Yes, and I had some problems with the format, it's not that I'm lazy :-)
I'll deal with that when I have power.
(ככה אומרים "כשיהיה לי כוח"?)

src/scrapers/mizrahi.js
baruchiro: My account is almost unused. I hope I can see what you said.
If not, maybe you know someone with an account at Mizrahi Bank, that we can write a code that will take out all the `json` and see how to detect such transaction?

baruchiro:
> I see that you are currently working on it. drop a message once you complete your changing.

Probably it will be at the weekend.

src/scrapers/mizrahi.js
baruchiro: I don't see any pending transactions.
I probably can't handle it. I will try to find someone with an account at Mizrahi Bank who will agree to run the code and check the JSON.

baruchiro:
@esakal Review please.

src/scrapers/mizrahi.js
baruchiro: Okay, I found "תנועות עתידיות". I'll handle it.

src/scrapers/mizrahi.js
esakal: Actually it is a good test case verifying the types of a transaction properties. Feel free to add it to Mizrachi tests and we can later add them to other scrapers

brafdlog:
Is this going to to be merged?

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,155 @@
...
+    mizrahixsrftoken: request.headers().mizrahixsrftoken,
+    'Content-Type': request.headers()['content-type'],
+  };
+}
+
+function convertTransactions(txns) {
+  return txns.map((row) => {
+    // row.MC02PeulaTaaEZ: 2019-09-10T00:00:00
+    const txnDate = moment(row.MC02PeulaTaaEZ, moment.HTML5_FMT.DATETIME_LOCAL_SECONDS)
+      .format(ISO_DATE_FORMAT);
```
eshaham: @baruchiro @esakal why not simply use moment's built-in function, `toISOString()`? we've been doing that across the repo :-)

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,155 @@
...
+function createHeadersFromRequest(request) {
+  return {
+    mizrahixsrftoken: request.headers().mizrahixsrftoken,
+    'Content-Type': request.headers()['content-type'],
+  };
+}
+
+function convertTransactions(txns) {
+  return txns.map((row) => {
+    // row.MC02PeulaTaaEZ: 2019-09-10T00:00:00
```
eshaham: @baruchiro can you please remove the code comment?

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,155 @@
...
+  ISO_DATE_FORMAT,
+} from '../constants';
+import { BaseScraperWithBrowser, LOGIN_RESULT } from './base-scraper-with-browser';
+import { fetchPostWithinPage } from '../helpers/fetch';
+import { waitForNavigation } from '../helpers/navigation';
+import { pageEvalAll } from '../helpers/elements-interactions';
+
+const BASE_URL = 'https://www.mizrahi-tefahot.co.il';
+const LOGIN_URL = `${BASE_URL}/he/bank/Pages/Default.aspx`;
+const AFTER_LOGIN_BASE_URL = /https:\/\/mto\.mizrahi-tefahot\.co\.il\/ngOnline\/index\.html#\/main\/uis/;
```
eshaham: @baruchiro why aren't we using a regular string here? would also be great to reuse `https://mto.mizrahi-tefahot.co.il` as something like `BASE_APP_URL`.

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,155 @@
...
+    this.navigateTo(OSH_PAGE, this.page);
+    this.request = await this.page.waitForRequest(TRANSACTIONS_REQUEST_URL);
+  }
+
+  getLoginOptions(credentials) {
+    return {
+      loginUrl: `${LOGIN_URL}`,
+      fields: createLoginFields(credentials),
+      submitButtonSelector: '#ctl00_PlaceHolderLogin_ctl00_Enter',
+      postAction: async () => this.sniffOshRequest(),
```
eshaham: @baruchiro I couldn't figure out why we're sniffing osh request during the `postAction`. Couldn't we simply move everything inside the `sniffOshRequest()` function to the beginning of the `fetchData()` function?
that way we don't need to have a `request` field in the class (unless I'm missing something).

src/scrapers/mizrahi.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async fetchData() {
+    const data = CreateDataFromRequest(this.request, this.options.startDate);
+    const headers = createHeadersFromRequest(this.request);
+
+    const response = await fetchPostWithinPage(this.page,
+      TRANSACTIONS_REQUEST_URL, data, headers);
+
+    if (response.header.success === false) {
```
eshaham: @baruchiro we prefer to have `if (!response.header.success)`

src/scrapers/mizrahi.js
baruchiro: I like to leave a comment with a real example when I handle formats.

baruchiro:
@eshaham @esakal 
What is the convention?
When someone opens a _Change Request_, is it the one who does _Resolve Conversion_ or can I resolve it too?

src/scrapers/mizrahi.js
baruchiro: > @baruchiro why aren't we using a regular string here? would also be great to reuse `https://mto.mizrahi-tefahot.co.il` as something like `BASE_APP_URL`.

@eshaham Did you notice that?

src/scrapers/mizrahi.js
baruchiro: Sorry, my fix is broken.
If something is wrong in our request, the response will be `response.header.success === false`.
But if everything is OK and the response contain the transaction, **the response will be `response.header.success === null`**.

src/scrapers/mizrahi.js
baruchiro: Resolved.
But there is a workaround for #277 .

src/scrapers/mizrahi.js
baruchiro: I mentioned you in a comment above, but I'm not sure what variable you mean?
I thought you meant `AFTER_LOGIN_BASE_URL`, but maybe you meant `BASE_URL` and/or `LOGIN_URL`?

esakal:
@baruchiro 
> What is the convention?
When someone opens a Change Request, is it the one who does Resolve Conversion or can I resolve it too?

This is a great question. imo the one who opens the change request should 'resolve' it, the one who fixes it should comment to it with 'done' or something similar. This is what we are doing in my workplace but I'm not sure what Elad think about it. 

I just want to note that generally I'm not a fan of github Code review support, in my work I prefer using more mature solution like https://reviewable.io/ unless it is very small PRs / libraries.

src/scrapers/mizrahi.js
eshaham: @baruchiro I still prefer not to keep it - I try to refrain from leaving code comments in this repo

src/scrapers/mizrahi.js
eshaham: @baruchiro sababa, but
1. you can also create regex strings with regular strings. I usually prefer to either do that, or if I feel this format is necessary, I only use it inside the `match` (or `replace`) functions. I dislike using a regex string in this format out of the context of its usage
2. you can still create the `BASE_APP_URL` constant, right?

src/scrapers/mizrahi.js
eshaham: @baruchiro no, I just referred to `https://mto.mizrahi-tefahot.co.il` and adding it to it's own constant (because it's also being used below 👇)

src/scrapers/mizrahi.js
eshaham: @baruchiro 😞 that's terrible. these banks and their awful API definitions. fine, let's keep it.

eshaham:
> @eshaham @esakal
> What is the convention?
> When someone opens a Change Request, is it the one who does Resolve Conversion or can I resolve it too?

yes, a good question indeed. what I usually try to do is:
1. have the contributer fix the comments and add a note after the comment saying fixed (I usually like to add a commit reference) or not fixed
2. the reviewer goes over these and resolves the conversation

src/scrapers/mizrahi.js
baruchiro: @eshaham Done.

src/scrapers/mizrahi.js
baruchiro: @eshaham Solved in https://github.com/eshaham/israeli-bank-scrapers/pull/256/commits/c8ad9fe388f90f45675e4fd3f1056e339e93f841, but I can't use the `BASE_APP_URL` in `AFTER_LOGIN_BASE_URL` because it must to be a `ReGex` object.

esakal:
@eshaham @baruchiro Hi, any chance wrapping this up so people will be able to use this scraper?

baruchiro:
I'm waiting for @eshaham  to confirm. I think.

src/scrapers/mizrahi.js
eshaham: 👍

Issue #146, opened 2018-07-16
CI support


lielran:
we can add Travis CI (free to open source projects) so we'll have default test runs + pr checks

eshaham:
@lielran we should probably add tests first... 😄 
running a CI just for linting isn't that beneficial IMO.
I've been wanting to add integration tests for a while now, but it's not that easy since we need to fetch and anonymize the data for each scraper. thoughts?

esakal:
I will let @eshaham to decide about this one. As using a CI is most effective as part of deployment process and he is the one that deploy this library.

Obviously if we had unit tests and would be relevant also to pull requests but as we are dealing with sensitive data and rely on data sources that are not in our control I'm not sure it would be easy to add unit tests.

Regarding the lint this, as @eshaham added [pre-commit](https://github.com/eshaham/israeli-bank-scrapers/blob/master/package.json#L18) enforcement we are covered in that area

@eshaham  @lielran  wdyt?

eshaham:
Let's add it only after we have a first test.
Closing this for now.

baruchiro:
> Let's add it only after we have a first test.

@eshaham Now is the time?

baruchiro:
Here are the stages of the proposal I am thinking about:
I present them as steps, since it's pretty obvious that it's all too much to implement at once. So each stage has value in itself (in my opinion).

1. **Code Owners:** Github allows us to enforce [Code Owners](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) by file, when Pull Request opens, if the file is modified, Code Owner will be set as a reviewer.
2. **Local Auto Test**: Instead of the Code Owner for each PR having to prepare an environment for the test (Clone, Checkout to PR merge, create test template file, run the test), there will be an automated script that will accept the PR number and a constant settings file, and run the test.
3. **Github API / Status Checks**: Send an automatic message (user-approved \ CI message (called "Status Check") when the local test passes.
4. **Dockerizing**.

PR #273, opened 2019-09-29
add prerelease workflow

esakal:
I admit that @baruchiro infected me with the github actions thing. I found a workflow that can be used to deploy a prerelease version to npm in a smart way and label it with `next` tag. 

@eshaham This might be really handy in our case. Having an automatic process that deploys a temporary `next` version will allow developers to use merged PRs changes immediately instead of waiting for an official release

I marked this PR as draft to consult with you both @eshaham and @baruchiro about the implementation. this workflow is using `epeli/npm-release` action. although it is very popular using others actions, I'm not sure how comfortable I feel to pass the NPM secret (of Elad) to that action. @baruchiro how about coping that action into our own project and using a local copy instead? its' license is MIT so I'm sure it is legit and I think we will feel more secure that way. wdyt? are you using somewhere local actions?

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Make prerelease to npm
-        uses: epeli/npm-release@v1
+        uses: epeli/npm-release@bc78fc8
         with:
           type: prerelease
           token: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,0 +1,15 @@
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  prerelease:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Make prerelease to npm
+        uses: epeli/npm-release@v1
+        with:
+          type: prerelease
+          token: ${{ secrets.NPM_TOKEN }}
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ If you are a developer and want to access the `next` version, install the librar
 npm install israeli-bank-scrapers@next --save
 ```  
 
-> Keep in mind that although this version should be stable as it passed our code review, it was deployed automatically using github action workflow without the usual tests we run manually done before we deploy the official version.  
+> Keep in mind that although this `next` version should be stable as it passed our code review, it was deployed automatically using github action workflow without the usual tests we run manually before we deploy the official version.  
 
 # `Israeli-bank-scrapers-core` library
 
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ The return value is a list of scraper metadata:
 # Getting deployed version of latest changes in master
 This library is currently deployed to NPM manually and not as part of automatic process. You should expect situations when code was pushed to master and wasn't deployed to NPM yet.
 
-If you are a developer and want to access the `next` version, install the library with `next tag as shown below:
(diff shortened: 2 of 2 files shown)


baruchiro:
First of all, I see that I lit the CI fire ;-)

Now, depending on your preferred way of working, you may want to consider first what system is right for us. In my PRs I use Github Actions, but maybe you want to use TravisCI?
I'm just asking if you would like to start a discussion. If so, you should open it at #146 .

-----
About using `epeli/npm-release`, I'm not familiar with the _prerelease_ pipeline, and also I'm not an expert in GithubActions, but I can say that yes, you can copy the `epeli/npm-release` to your repository.

But I think also there is a way to select a specific version of `epeli/npm-release`. So, if you checked the code in the current version of external action, you can use that version and even if the author will update the Action with bad code, you still use the verified version (but we should check if the author can override a specific version).

esakal:
@baruchiro, CI is something that is in this library roadmap yet adding unit-tests doesn't necessarily serve our needed since the companies endpoints are not predicted. Sure we can mock things and use snapshots, but most of the time it will break due to changes in the endpoints and not as a result of wrong code modifications. So we actually needs end-to-end testing which also has its' issues.

The reason I'm pointing the unit-test / end-to-end thing is that we are not pushing the CI forward till we will decide how to tackle the automatic tests. Without them CI server cannot actually publish to NPM and can cover very limited scenarios. The reason I added this PR is that I actually think that we can benefit from this feature immediately even without tackling the tests issue.

> In my PRs I use Github Actions, but maybe you want to use TravisCI?

Wether using Github Actions or TravisCI is in my opinion something that should be decided by @eshaham. Both will do the trick.

> About using epeli/npm-release, I'm not familiar with the prerelease pipeline, and also I'm not an expert in GithubActions, but I can say that yes, you can copy the epeli/npm-release to your repository.

can you guide me to a good reference or documentation that is using the latest github actions api? I found some examples that were using the former api with `main.workflows` prior to the YAML support.

baruchiro:
https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses

```yaml
steps:    
  - uses: actions/setup-node@74bc508 # Reference a specific commit
  - uses: actions/setup-node@v1      # Reference the major version of a release   
  - uses: actions/setup-node@v1.2    # Reference a minor version of a release  
  - uses: actions/setup-node@master  # Reference a branch
```

esakal:
@baruchiro thanks for the reference, I modified this PR to reference a specific commit. 

---

@eshaham there are two options:
1. using a specific commit as done in this PR (latest tag [epeli/npm-release@v1.1.1](https://github.com/epeli/npm-release/releases/tag/v1.1.1)).
2. coping the action into our repo and using relative path [as mentioned here](https://help.github.com/en/articles/workflow-syntax-for-github-actions#example-using-action-in-the-same-repository-as-the-workflow). 

I think option one is safe enough based on github actions documentation. see [here](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses):
>  ... Using the commit SHA of a released action version is the safest for stability and security. ...

## How to continue with this PR
1. review [epeli/npm-release@v1.1.1 source](https://github.com/epeli/npm-release/blob/bc78fc8caa52fa9541f83b3fbaa2fc9938bdf0ee/src/main.ts)
2. add to github actions `NPM_TOKEN` [see documentation here](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables)

eshaham:
@esakal I like this, let's try it 😄

eshaham:
@esakal sweet, seems to be working 😄

esakal:
Yesh! this will reduce the pressure of deploying new official commits after every merge :)

PR #290, opened 2019-10-27
Fix linebreak lint error on windows

### LINKED ISSUE #252: Remove Line Breaking validation from lint
Hi, in the `pre-commit` we run the `lint`, who is required to the `linebreaks to be 'LF'`.

But in Windows we work with `CRLF`, so we get the error: `Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style`.

Anyway Git commits all files as `LF`.

baruchiro:
close #252


esakal:
@baruchiro did you see https://stackoverflow.com/a/9977954 ? they seems to provide a bit more complex solution, what do you think about it?

 @eshaham are you experienced with something like that?

baruchiro:
Look, _Airbnb_ decided that all source files should be **LF**, and they put it in their `lint` base.

I don't know why this is required, but it causes interruptions in **Windows** work. I don't think it is too much to deal with, especially if we run `lint` on **Linux** at the end, so if someone forcefully put a file in **CRLF** to _Git_ (although _Git_ always saves **LF**), we will find out.

In my new GUI project (coming soon...) I set this line in my `.eslintsrc.js`:
``` node
'linebreak-style': process.platform === "win32"? 0: 2
```

esakal:
I'm favoring what you suggested to trick the eslint rule.
```
'linebreak-style': process.platform === "win32"? 0: 2
```

@eshaham do you agree?

eshaham:
@baruchiro @esakal sounds good to me

baruchiro:
@esakal I moved to `.eslintrc.js`. Please run the checks after you merge https://github.com/eshaham/israeli-bank-scrapers/pull/291 because in my machine the lint start to fail after my change.

baruchiro:
Closed. #291

PR #291, opened 2019-10-27
add cross-env dependency

esakal:
# Motivation
The new CI exposed two issues in windows:
1. the CRLF thing (handled in separated ticket) 
2. environment variables that are used in a non-crossed environment way. This PR address that one.

this PR now handle both issues (I merged #290) into this PR

@baruchiro fyi

image

closes #290


eshaham:
@esakal I'm kind of confused between this one and #290, as tests on both seem to be failing.
which one are we trying to merge first?

esakal:
I need to handle them both in the same PR, but it currently fails, I need to try and understand why.

baruchiro:
@esakal I see you took my PR.
Can I cancel the #290 ?

esakal:
@baruchiro @eshaham this one was very tricky, I even installed git+node on my wife windows just to be sure it happens only in circle CI. 

So the error actually happens as a result of a linting process that was triggered just before the test. We don't actually need it as we have a separate workflow that lint our PRs.

The tests seems to pass now, @eshaham please review and re-approve

PR #308, opened 2019-11-16
Support bankhapoalim be online

orzarchi:
Hi there!
Bank hapoalim be online is used for digital-only accounts, but it was unsupported until now.
The be online portal uses the exact same scraping procedure, with slightly altered urls.
This PR moves the poalim base url and some methods inside the scraper classes,  and then creates a simple override for the new scraper.


src/scrapers/hapoalim.js
```diff
@@ -142,18 +91,84 @@ function createLoginFields(credentials) {
...
+      possibleResults: this.getPossibleLoginResults(),
     };
   }
 
   async fetchData() {
-    return fetchAccountData(this.page, this.options, (msg) => this.notify(msg));
+    return this.fetchAccountData(this.page, this.options, (msg) => this.notify(msg));
+  }
+
+  async fetchAccountData(page, options) {
```
esakal: @orzarchi instead of moving those methods into the class, can you move them back and adjust the method signature instead. for example:
```
async fetchAccountData(page, baseUrl, options) {
...
}
```

orzarchi:
@esakal Fixed all comments

esakal:
@orzarchi we had an issue with the automatic tests in windows machines. It is now fixed in master. can you please pull master so your PR will re-run those tests? 

I assume @eshaham will approve and merge your branch soon.

Thanks again!

PR #309, opened 2019-11-17
Bump node version to 8.16.2

eshaham:
eslint 6 and above requires to run node version >= 8.10


eshaham:
@esakal yes - the trigger for the upgrade was an eslint 6 requirement, but there's no reason not to upgrade node to the latest 8 version

Issue #322, opened 2019-12-03
Discount scraper stopped working


brafdlog:
Discount scraper fails with the following error: `Login failed with UNKNOWN_ERROR error`

brafdlog:
The issues is that the expected success url after login is: `https://start.telebank.co.il/apollo/core/templates/default/masterPage.html#/MY_ACCOUNT_HOMEPAGE`

But the actual url is `https://start.telebank.co.il/apollo/core/templates/RETAIL/masterPage.html#/MY_ACCOUNT_HOMEPAGE`

I see that discount have updated their ui and I assume in this new ui the success url changed

brafdlog:
Created a pr to fix #323

esakal:
pr #323 merged, as mentioned in the PR It might take time until it will be released as official latest. in the meanwhile feel free to pull the `next` tag from npm:
```
npm i israeli-bank-scrapers@next
```

PR #323, opened 2019-12-03
Fix success url for discount bank's new ui #322

brafdlog:
Discount updated their ui and now after login the url is: 

`https://start.telebank.co.il/apollo/core/templates/RETAIL/masterPage.html#/MY_ACCOUNT_HOMEPAGE` 

as opposed to:

 
`https://start.telebank.co.il/apollo/core/templates/default/masterPage.html#/MY_ACCOUNT_HOMEPAGE`

### DIFF
--- a/src/scrapers/discount.js
+++ b/src/scrapers/discount.js
@@ -79,7 +79,7 @@ async function navigateOrErrorLabel(page) {
 
 function getPossibleLoginResults() {
   const urls = {};
-  urls[LOGIN_RESULT.SUCCESS] = [`${BASE_URL}/apollo/core/templates/default/masterPage.html#/MY_ACCOUNT_HOMEPAGE`];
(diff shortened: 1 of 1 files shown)


brafdlog:
@esakal Great.
If we see they segment the users we can add both urls to the array.

brafdlog:
@esakal Can you publish a new npm version?

esakal:
@brafdlog Hi. Unfortunately I don't have permission to publish to NPM.  Only @eshaham has the relevant permissions in NPM. 

this is why we added the github actions that publish next version automatically once a commit reaches `master` so people will be able to use them asap.

I know that you wanted to wait to the official version but  I believe you can consider `next` as the official as in this repo we don't merge to master anything that will not reach the official version anyway.

eshaham:
@brafdlog done

PR #336, opened 2019-12-20
Change Isracard default start date to earlier

Urigo:
Isracard is capable of returning older history then just 1 year.

On my account I was able to go back 4 years but maybe for others it might be even longer.

Not sure if it's a known thing and you just wanted to send back 1 year as default?
For me it made me think that the scraper can't bring earlier data but I decide to try it anyway.

### DIFF
--- a/src/scrapers/base-isracard-amex.js
+++ b/src/scrapers/base-isracard-amex.js
@@ -268,7 +268,7 @@ class IsracardAmexBaseScraper extends BaseScraperWithBrowser {
   }
 
   async fetchData() {
-    const defaultStartMoment = moment().subtract(1, 'years');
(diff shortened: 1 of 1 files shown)


esakal:
@Urigo Hi,
This is the design of the library. the default 1 year value is cross all scrapers and should remain like that. 

But we are actually using it for two different purposes - default value and max value. I agree that we should consider not limiting the max value to match the default value. 

So I think you should adjust the code: revert default value to 1 year and add max default value which will be 4 years.

But in order to do that, we need to check if a common user can fetch transactions of 4 years from the application. since we simulate regular users we don't want to abuse the api and allow things that are not permitted to the users imo.

@Urigo can you check if 4 years is valid value in their site?

Issue #341, opened 2020-01-06
Leumi Card renamed to MAX


baruchiro:
Just see in the URLs and so...

adiohana:
Hi @baruchiro 
I would like to contribute to this project.
any chance for some more details about this issue?

baruchiro:
This project uses the old name of "LeumiCard", and "LeumiCard" changed their name to "Max", so you need to change all "Leumi Card" in this project to "Max".

I'm just waiting to hear @eshaham 's opinion, but I think you can start because I don't see a reason for it to be a problem.

eshaham:
> I'm just waiting to hear @eshaham 's opinion, but I think you can start because I don't see a reason for it to be a problem.

Totally makes sense

> Hi @baruchiro
I would like to contribute to this project.
any chance for some more details about this issue?

PR welcomed

adiohana:
@eshaham @baruchiro 
Just created my first PR :]
I have a strong background in financial web-scraping (mainly in US market) and I just heard about this cool project; I will definitely contribute more in the future.

eshaham:
merged 👍

PR #348, opened 2020-01-13
Renamed Leumi Card to Max

adiohana:
Context: Leumi Card recently made a re-branding and changed their name to "Max". 

Changes:
- Refactor Leumi Card to Max in all places

Related to #341


src/scrapers/factory.js
```diff
@@ -25,8 +25,8 @@ export default function createScraper(options) {
       return new OtsarHahayalScraper(options);
     case 'visaCal':
       return new VisaCalScraper(options);
-    case 'leumiCard':
-      return new LeumiCardScraper(options);
+    case 'max':
```
baruchiro: Why not keeping the `leumiCard` case, and printing a deprecation warning?

adiohana: @baruchiro 
you mean only keep the "leumiCard" case for backward compatibility (with deprecation warning)? and create a new case with "max"?

baruchiro: Yes, I don't know if the next syntax is correct, but something like this:

```node
case 'leumiCard':
  console.log(...);
case 'max':
  return new LeumiCardScraper(options);
```

baruchiro: So the `case 'leumiCard'` will return `LeumiCardScraper`

baruchiro:
Consider to replace the `LeumiCardScraper` class name.

src/scrapers/factory.js
adiohana: thanks.
since the COMPANY_ID is hard coded in the test files, we will have to find a work around for this.

adiohana: @baruchiro 
made the necessary changes and kept the 'leumiCard' option.
all tests are looking good.
let me know what you think

baruchiro: In my understanding, the project reveals two things:
A list of the **Scrapers**, and the **Factory**.

I think if someone wrote the code based on the **Scrapers** list, if you rename one of them, their code wouldn't change, because it would automatically fit.

The problem is if someone uses the **Factory** directly by sending a string with the value `leumiCard`. In this case, I think it's still worth supporting.

Therefore, I suggest that you remove any reference to `LeumiCard`, from both the **Scrapers** list (then you will need to delete the unnecessary file `leumi-card.test.js`) and the **README** file, and the only reference you leave is in the **Factory**.

@eshaham That's my opinion, but I think we might both be right. Waiting for your decision.

adiohana: @eshaham can we get your input here?

src/definitions.js
```diff
@@ -29,6 +29,10 @@ export const SCRAPERS = {
     name: 'Leumi Card',
     loginFields: ['username', PASSWORD_FIELD],
   },
+  max: {
```
eshaham: @adiohana we should have a todo above leumi card to remove it in a major version

src/scrapers/factory.js
eshaham: @adiohana @baruchiro we should remove the test file for sure.
for backward compatibility, I like the approach above, of keeping support for `leumiCard` but printing a warning.
we can do that until we reach a major version.

src/scrapers/max.test.js
```diff
@@ -0,0 +1,52 @@
...
+    const scraper = new MaxScraper(options);
+    const result = await scraper.scrape(testsConfig.credentials.max);
+    expect(result).toBeDefined();
+    const error = `${result.errorType || ''} ${result.errorMessage || ''}`.trim();
+    expect(error).toBe('');
+    expect(result.success).toBeTruthy();
+
+    exportTransactions(COMPANY_ID, result.accounts || []);
+  });
+});
```
eshaham: @adiohana given my comment above about remove the leumi card test, I'd rather see a file rename here (plus some text changes), rather than an entirely new file

tests/.tests-config.tpl.js
```diff
@@ -14,7 +14,8 @@ export default {
     // leumi: { username: '', password: '' },
     // discount: { id: '', password: '', num: '' },
     // otsarHahayal: { username: '', password: '' },
-    // leumiCard: { username: '', password: '' },
+    // leumiCard: { username: '', password: '' }, //Deprecated - use 'max' instead
```
eshaham: @adiohana remove

adiohana:
@eshaham 
made all the changes per your request, thanks for the feedback.

src/definitions.js
baruchiro: Now I have an annoying problem.
I present a list of scrapers, which is based on `SCRAPERS`. The user enters and saves details of the scraper, along with his key (`leumiCard`) so that in the next use he will use the key to get an instance of the scraper.

So to be ready for change, I have a migration to change the key from `leumiCard` to `max`.
But it won't help me, because as long as `leumiCard` still exists in `SCRAPERS`, the user will continue to create new `leumiCard` keys.

PR #369, opened 2020-02-06
issue 365 -  Mizrahi fetchData request will happen during postAction …

SafiX:
Scraping would timeout waiting for TRANSACTIONS_REQUEST_URL request since this request already finished during the postAction.

### DIFF
--- a/src/scrapers/mizrahi.js
+++ b/src/scrapers/mizrahi.js
@@ -104,16 +104,18 @@ class MizrahiScraper extends BaseScraperWithBrowser {
       loginUrl: `${LOGIN_URL}`,
       fields: createLoginFields(credentials),
       submitButtonSelector: '#ctl00_PlaceHolderLogin_ctl00_Enter',
-      postAction: async () => waitForNavigation(this.page, { waitUntil: 'networkidle0' }),
+      postAction: async () => {
+        this.request = await this.page.waitForRequest(TRANSACTIONS_REQUEST_URL);
+        waitForNavigation(this.page, { waitUntil: 'networkidle2' });
+      },
       possibleResults: getPossibleLoginResults(),
     };
   }
 
   async fetchData() {
     await this.navigateTo(OSH_PAGE, this.page);
-    const request = await this.page.waitForRequest(TRANSACTIONS_REQUEST_URL);
-    const data = CreateDataFromRequest(request, this.options.startDate);
-    const headers = createHeadersFromRequest(request);
+    const data = CreateDataFromRequest(this.request, this.options.startDate);
+    const headers = createHeadersFromRequest(this.request);
 
     const response = await fetchPostWithinPage(this.page,
       TRANSACTIONS_REQUEST_URL, data, headers);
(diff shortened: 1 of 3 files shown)


baruchiro:
I think there was a [reason](https://github.com/eshaham/israeli-bank-scrapers/pull/256#discussion_r331409249) for that, but I don't remember anymore.

esakal:
@baruchiro I assume you have Mizrachi account. Does it work for you without this PR? does it work for you with this PR?

baruchiro:
I assume @SafiX have a Mizrahi account too.

But hopefully I'll check it out this weekend.

esakal:
Obviously, but sometimes the flow is different between account types or content. If you manage to check it later this weekend let us know. If not I will merge it Tomorrow evening

src/scrapers/mizrahi.js
```diff
@@ -104,16 +104,18 @@ class MizrahiScraper extends BaseScraperWithBrowser {
...
+      postAction: async () => {
+        this.request = await this.page.waitForRequest(TRANSACTIONS_REQUEST_URL);
+        waitForNavigation(this.page, { waitUntil: 'networkidle2' });
+      },
       possibleResults: getPossibleLoginResults(),
     };
   }
 
   async fetchData() {
     await this.navigateTo(OSH_PAGE, this.page);
```
baruchiro: I tested it and now I remember why it is not in the `postAction`. Do you see the `await this.navigateTo(OSH_PAGE, this.page);`?
The `OSH_PAGE` triggers the request to `TRANSACTIONS_REQUEST_URL`. In Mizrahi, you can select your default page for post login. If you select the "הודעות מהבנקאי" page, the `TRANSACTIONS_REQUEST_URL` will not triggered until you `navigateTo(OSH_PAGE)`,

baruchiro: @eshaham @esakal I think I don't approve this PR.

@SafiX Did you have a problem that you opened this PR for?
Because as far I remember, the `waitForRequest` will take the request even if it recorded before the `waitForRequest` command.

SafiX: @baruchiro - yes, with my account which I have the default "default page for post login" the scraper simply doesn't work. The test for it fails as well.

From what I  understand, my PR doesn't work with your configuration?

I guess the solution should have both of our bank ''default page for post login" configuration work.

I'll push a fix to the branch which will handle both cases.

SafiX: @baruchiro - Check out my latest commit to this branch, it should work for both our cases now.

SafiX: @baruchiro - 
ok just FYI, I went to - "הגדרת עמוד כניסה"
And saw that mine was - " פורטל עו''ש "

When I changed it to "תיבת ההודעות" the scraper was working again.

So I think that some people don't change their default post login page, I tested and saw that now, with the additional fix, it will work with both cases.

Would you be able to test as well on your side?

SafiX:
So after changing the default landing page to my account, I can't reconstruct the original issue anymore.

baruchiro:
I haven't tested the case yet, but I think it should fit for all landing
pages, regardless of user modification.

SafiX:
Yes, I agree, however I can't reconstruct it anymore.

PR #412, opened 2020-04-11
Beinleumi contribution

dudiventura:
Hi,
I've helped with the contribution of the Union Bank scraper to this project along with #intuit team and I saw that you're missing the Beinleumi Bank as well.

Would love to offer this script as another contribution for this project effort :)

### DIFF
--- a/src/scrapers/beinleumi.js
+++ b/src/scrapers/beinleumi.js
@@ -15,8 +15,10 @@ const LOGIN_URL = `${BASE_URL}/MatafLoginService/MatafLoginServlet?bankId=FIBIPO
 const TRANSACTIONS_URL = `${BASE_URL}/wps/myportal/FibiMenu/Online/OnAccountMngment/OnBalanceTrans/PrivateAccountFlow`;
 const DATE_FORMAT = 'DD/MM/YYYY';
 const NO_TRANSACTION_IN_DATE_RANGE_TEXT = 'לא נמצאו נתונים בנושא המבוקש';
-const DATE_COLUMN_CLASS = 'date first';
-const DESCRIPTION_COLUMN_CLASS = 'reference wrap_normal';
+const DATE_COLUMN_CLASS_COMPLETED = 'date first';
+const DATE_COLUMN_CLASS_PENDING = 'first date';
+const DESCRIPTION_COLUMN_CLASS_COMPLETED = 'reference wrap_normal';
+const DESCRIPTION_COLUMN_CLASS_PENDING = 'details wrap_normal';
 const REFERENCE_COLUMN_CLASS = 'details';
 const DEBIT_COLUMN_CLASS = 'debit';
 const CREDIT_COLUMN_CLASS = 'credit';
@@ -24,8 +26,8 @@ const ERROR_MESSAGE_CLASS = 'NO_DATA';
 const ACCOUNTS_NUMBER = 'div.fibi_account span.acc_num';
 const CLOSE_SEARCH_BY_DATES_BUTTON_CLASS = 'ui-datepicker-close';
 const SHOW_SEARCH_BY_DATES_BUTTON_VALUE = 'הצג';
-const TRANSACTIONS_TABLE = 'table[id*=\'dataTable\']';
-const TRANSACTIONS_PAGINATION_LIMIT = 20;
+const COMPLETED_TRANSACTIONS_TABLE = 'table#dataTable077';
+const PENDING_TRANSACTIONS_TABLE = 'table#dataTable023';
 const NEXT_PAGE_LINK = 'a#Npage.paging';
 
 function getPossibleLoginResults() {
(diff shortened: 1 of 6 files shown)


src/scrapers/beinleumi.js
```diff
@@ -0,0 +1,265 @@
...
+
+async function navigateToNextPage(page) {
+  await clickButton(page, NEXT_PAGE_LINK);
+  await waitForNavigation(page);
+}
+
+async function scrapeTransactions(page) {
+  const txns = [];
+  for (let i = 0; i < TRANSACTIONS_PAGINATION_LIMIT; i += 1) {
+    txns.push(...await extractTransactions(page));
```
eshaham, line 186: @dudiventura could you extract to a variable?
```suggestion
    const transactions = await extractTransactions(page);
    txns.push(...transactions);
```

src/scrapers/beinleumi.js
```diff
@@ -0,0 +1,265 @@
...
+    try {
+      await navigateToNextPage(page);
+    } catch (error) {
+      if (error.message.includes(CLICK_ELEMENT_NOT_EXISTS_ERROR_MSG)) {
```
eshaham, line 190: @dudiventura could you explain why you're breaking based on this error?

src/scrapers/beinleumi.test.js
```diff
@@ -0,0 +1,52 @@
...
+    extendAsyncTimeout(); // The default timeout is 5 seconds per async test, this function extends the timeout value
+  });
+
+  test('should expose login fields in scrapers constant', () => {
+    expect(SCRAPERS.beinleumi).toBeDefined();
+    expect(SCRAPERS.beinleumi.loginFields).toContain('username');
+    expect(SCRAPERS.beinleumi.loginFields).toContain('password');
+  });
+
+  maybeTestCompanyAPI(COMPANY_ID, (config) => config.companyAPI.invalidPassword)('should fail on invalid user/password"', async () => {
```
eshaham, line 22: @dudiventura seems like you have redundant quotes sign
```suggestion
  maybeTestCompanyAPI(COMPANY_ID, (config) => config.companyAPI.invalidPassword)('should fail on invalid user/password', async () => {
```

src/scrapers/beinleumi.js
dudiventura: Hi @eshaham,
Yes, I've tried first to use the function **elementPresentOnPage** to check if the "next page" link element presents in the page.
But for some reason I kept getting a positive answer (true) from the function even when the element wasn't there.

And when the script tries to click on element using the function **clickButton** while the element doesn't exists I get the error "Cannot read property".

So when this error is thrown, I catch it, validate the this is the error and by that assume that the element doesn't exists in the page and there is no next page so I can break the loop there.

I know it's not the most elegant way, so if you have a suggestion to another way I'll be happy to learn something new.

src/scrapers/beinleumi.test.js
dudiventura: You're right,
Will delete it and push on the next commit.

src/scrapers/beinleumi.js
dudiventura: I agree, it will look cleaner that way.
Will do that and push on the next commit.

src/scrapers/beinleumi.js
esakal: hmm... the method `elementPresentOnPage` should be reliable. just to make sure you have tried:
1. pausing the browser and making sure the selector is correct.
```
await page.evaluate(() => {
  debugger;
});
```
2. calling `elementPresentOnPage` with await.

src/scrapers/beinleumi.js
```diff
@@ -0,0 +1,265 @@
...
+}
+
+function getTransactionCredit(tds, transactionsColsTypes) {
+  return (tds[transactionsColsTypes[CREDIT_COLUMN_CLASS]] || '').trim();
+}
+
+function extractTransactionDetails(txnRow, transactionsColsTypes) {
+  const tds = txnRow.innerTds;
+  return {
+    status: TRANSACTION_STATUS.COMPLETED,
```
esakal, line 99: it seems that you are always returning status `COMPLETED`. are you familiar with `PENDING` status? it usually happens on holidays and weekends. Do you know if BeinLeumi has pending transactions?

dudiventura: I'm not familiar with PENDING transactions on the Beinleumi and I think they show completed transactions only.

src/scrapers/beinleumi.js
dudiventura: Thank you for your comment, I'm checking it again.

dudiventura: Hi @esakal,
You are very right, I had a missing "await" in the code and therefore I didn't get the correct answer from the function.
I'm deploying a fix now.

src/scrapers/beinleumi.js
esakal: I think all banks show PENDING transactions, try to check your account on Tuesday / Wednesday. maybe will will get lucky and you will have pending transactions then :)

esakal:
@eshaham I approved the PR, beside the Pending transactions question I have no other comments. It is not easy to see Pending transactions because they are shown only on specific circumstances.  Do you think we should leave it for now?

src/scrapers/beinleumi.js
dudiventura: @eshaham  and @esakal , 
Please **do not** approve this PR yet.

I've decided to investigate Eran's assumption on the bank website, so I've made a money movement between my accounts (from different banks).
I've fountd out that the transaction of the money movement I've just made appears on the bank website in another table under the title ״תנועות מהיום״.

The problem in the source code is that both of the transactions tables have non meaningful ids (dataTable077 and dataTable023) and it just means that **I have a bug** in my code which now extract transactions from both tables and set them as COMPLETED transactions.

Please allow me to fix that bug first, and to support PENDING transactions as well.

src/scrapers/beinleumi.js
```diff
@@ -0,0 +1,268 @@
...
+  for (let i = 0; i < TRANSACTIONS_PAGINATION_LIMIT; i += 1) {
+    const currentPageTxns = await extractTransactions(page);
+    txns.push(...currentPageTxns);
+    hasNextPage = await checkIfHasNextPage(page);
+    if (hasNextPage) {
+      await navigateToNextPage(page);
+    } else {
+      break;
+    }
+  }
```
eshaham, line 198: @dudiventura now that I better understand this loop, wouldn't it be simpler (and cleaner) to do something like this?
```suggestion
  let hasNextPage = true;
  while (hasNextPage) {
    const currentPageTxns = await extractTransactions(page);
    txns.push(...currentPageTxns);
    hasNextPage = await checkIfHasNextPage(page);
    if (hasNextPage) {
      await navigateToNextPage(page);
    }
  }
```

dudiventura: Changed to do-while loop

src/scrapers/beinleumi.js
dudiventura: Hi @eshaham and @esakal,
I've added a support of pending transactions on the last commit and I think the script is ready now.
Will be happy to hear your opinion.

PR #404, opened 2020-04-03
Typescript

esakal:
# Goals
Use Typescript with israeli-bank-scrapers library.

@eshaham since this PR is going to have several iterations, please review and comment using reviewable (use the button at the bottom of this description to redirect to reviewable)

# Motivation
As we are looking toward making this library extendible and customizable, we expect the library api to grow and we would like to use Typescript to help us to scale.
[read more](https://medium.com/@jtomaszewski/why-typescript-is-the-best-way-to-write-front-end-in-2019-feb855f9b164)

# Tasks
- [x] add Typescript dependencies & configuration.
- [x] use Babel & Typescript during build process.
- [x] convert all helpers to Typescript.
- [x] convert constants to Typescript and use enums.
- [x] add core type declaration for Transaction, LoginResult, and others.
- [x] update (jest) tests infrastructure to use Typescript.
- [x] convert Leumi account to Typescript.
- [x] fix lint warnings.
- [x] remove test folder as part of the building process
- [x] convert all remaining files from js to ts.
- [x] fix type of LoginOptions > possibleResults
- [x] run local test on available companies
- [x] update documentation.
- [x] make sure external apps works with this library.
- [x] make sure external apps works with this library **in core mode**.
- [x] ask developers to test their companies as well.

closes #152
closes #487 

---
This change is [Reviewable](https://reviewable.io/reviews/eshaham/israeli-bank-scrapers/404)

PR #464, opened 2020-06-22
add git ignore file as preparation for typescript branch

esakal:
# Motivation

* Until people will adapt Typescript (#404) they might want to experiment with it. when they will go back to master they might accidentally commit there credentials test file as we needed to move the `tests` folder to `src/tests` and when switching to master the `.gitignore` file is removed from that folder

see the reproducing below, note how the file is marked to commit (visualized in my IDE with red color)
image

once this PR will be in master, the file will be marked as ignored (visualized in my IDE with gray color) 
image


src/tests/.gitignore
```diff
@@ -0,0 +1,3 @@
+.*.*
+.*.*
```
eshaham, line 2: can't believe I'm actually commenting on this PR, but these 2 lines are identical

PR #469, opened 2020-06-26
support custom args provided by user to puppeteer lunch

### LINKED ISSUE #419: Passing args to puppeteer launch method
In my project, I run the scrapper from inside a docker container, and I want to make is as easy as possible. To prevent some issues that come up on the container and puppeteer, I need to pass an argument to the underline chrome process `--no-sandbox`.
I do not see any option to pass it through the api currently.
I guess that the required code change would be under https://github.com/eshaham/israeli-bank-scrapers/blob/master/src/scrapers/base-scraper-with-browser.js, when calling `puppeteer.launch` to pass an args variable.
I would be happy to work on a solution if you would direct me to how do you think the args should be passed from the api.

esakal:
# Overview
Add optional args option to be provided as-is to the `puppeteer.lunch` command. Includes unit-tests.

#closes #419

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,8 @@ copy and adjust the json below with relevant credentials and assign it to enviro
      "startDate": "2019-06-01",
      "combineInstallments": false,
      "showBrowser": true,
-     "verbose": false
+     "verbose": false,
+     "args": []
    },
    "credentials": {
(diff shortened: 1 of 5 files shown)


esakal:
@eshaham wdyt?

esakal:
@eshaham Hi!
any update about this one?

PR #473, opened 2020-06-29
Added message for blocked account on visaCal

talsalmona:
After a few failed login attempts, Visa Cal will ask you to change your password.
This will spit out a message for this use case instead of an unknown error.


package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">= 5.3.0"
   },
   "main": "lib/index.js",
+  "publishConfig": { "registry": "https://npm.pkg.github.com/" },
```
eshaham, line 10: @talsalmona no need for this change

src/constants.js
```diff
@@ -24,6 +24,7 @@ export const LOGIN_RESULT = {
   SUCCESS: 'SUCCESS',
   INVALID_PASSWORD: 'INVALID_PASSWORD',
   CHANGE_PASSWORD: 'CHANGE_PASSWORD',
+  BLOCKED_USER: 'BLOCKED_USER',
```
eshaham, line 27: @talsalmona sorry for being picky, I think that `USER_BLOCKED` works better as a login result in this case

esakal:
@talsalmona Hello :)

In readme.md file there is a place that list the available results and in contributing.md there are few.

can you please add the new option to that list, thanks!

package.json
talsalmona: Sorry, was unintentional. I'll remove it.

esakal:
@talsalmona Hi, there are two more comments open and we will be able to merge it.
Thanks :)

esakal:
@eshaham how do you want to proceed with this one?

talsalmona:
Hi, sorry for the delay. Last week was very busy.
Please review my latest changes.
Also, I believe the README and CONTRIBUTING had an error where the possible return type on error we not correct.
WDYT?

esakal:
@eshaham likewise :) please release a new official version when possible.

@talsalmona In the meanwhile feel free to pull the next tag from npm:

```
```

PR #483, opened 2020-07-18
Solve AMEX scraping issue by upgrading puppeteer to version 5

### LINKED ISSUE #482: Cannot scrape American express with Puppeteer version ^2.0.0
When trying to scrape Amex a reset is returned from the network
```
"GENERIC net::ERR_CONNECTION_RESET at https://he.americanexpress.co.il/personalarea/Login"
```

To solve it we need to use updated chromium that is provided in Puppeteer 5.x.

esakal:
see #482 

closes #482


esakal:
@eshaham this PR upgrades puppeteer from 2.x to 5.x.
Puppeteer 5.x requires node version 10.x as seen [here](https://github.com/puppeteer/puppeteer/blob/1d08e274a85812d7b88dd0b72e5e26547430c540/package.json#L8)

Is this a problem for us? I don't think we really have a choice as we are way behind the latest puppeteer version.

Regarding the automatic checks. As you can see once I removed node 8.x from the tests everything passes. The only reason it is not marked as completed is that Github waits for the build on 8.x to complete but now we don't run them at all so it will never complete. But the other relevant tests do complete successfully.

esakal:
@eshaham how do you want to proceed with this one?

esakal:
@eshaham I merged master (with the Typescript support) into this PR so it should be ready to merge. 
Let's release it as part of version 1.0.0 as you suggested?

Let me know if I can merge it

Just a reminder that the relevant tests passed, only those marked for node v8.0 which is not relevant any more causes the status to appear incomplete

eshaham:
@esakal yes, let's do it 👍

esakal:
Hmm... I don't have administrative permissions to bypass the ghost statuses thing. @eshaham can you please merge it?

esakal:
@eshaham can you handle this one? I cannot merge it on my own as we have orphaned tests

eshaham:
@esakal done

PR #490, opened 2020-07-30
Fix Mizrahi

### LINKED ISSUE #456: Mizrahi scraper does not work in v0.9.0
hey @baruchiro . I see you're Mizrahi supporter.

I have a valid Mizrahi account. I'm trying to work with Mizrahi scraper. And it looks like broken.
I tested through [israeli-ynab-updater](https://github.com/eshaham/israeli-ynab-updater) and    mizrahi.test.js (israeli-bank-scrapers)

I'm using version - v0.9.0

I have the following errors

In israeli-ynab-updater :
```
scraping Mizrahi Bank
Mizrahi Bank: START_SCRAPING
Mizrahi Bank: INITIALIZING
Mizrahi Bank: LOGGING_IN
Mizrahi Bank: LOGIN_SUCCESS
Mizrahi Bank: TERMINATING
Mizrahi Bank: END_SCRAPING
success: false
error type: GENERIC
error: Timeout exceeded while waiting for event
```

In Mizrahi Tests resutls:
image

I will appreciate if you can help with that.

Thanks

baruchiro:
close #456

### DIFF
--- a/src/helpers/elements-interactions.js
+++ b/src/helpers/elements-interactions.js
@@ -2,10 +2,6 @@ async function waitUntilElementFound(page, elementSelector, onlyVisible = false,
   await page.waitForSelector(elementSelector, { visible: onlyVisible, timeout });
 }
 
-async function waitForText(page, text, elementSelector = 'body', timeout = 30000) {
-  await page.waitForFunction(`document.querySelector('${elementSelector}') && document.querySelector('${elementSelector}').innerText.includes('${text}')`, { timeout });
-}
-
 async function fillInput(page, inputSelector, inputValue) {
   await page.$eval(inputSelector, (input) => {
     const inputElement = input;
@@ -60,7 +56,6 @@ async function dropdownElements(page, selector) {
 
 export {
   waitUntilElementFound,
-  waitForText,
   fillInput,
   clickButton,
   clickLink,
--- a/src/helpers/elements-interactions.js
+++ b/src/helpers/elements-interactions.js
@@ -2,6 +2,10 @@ async function waitUntilElementFound(page, elementSelector, onlyVisible = false,
   await page.waitForSelector(elementSelector, { visible: onlyVisible, timeout });
 }
 
+async function waitForText(page, text, elementSelector = 'body', timeout = 30000) {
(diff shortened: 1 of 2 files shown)


src/helpers/elements-interactions.js
```diff
@@ -11,8 +11,7 @@ async function fillInput(page, inputSelector, inputValue) {
 }
 
 async function clickButton(page, buttonSelector) {
-  const button = await page.$(buttonSelector);
-  await button.click();
+  await page.$eval(buttonSelector, (el) => el.click());
```
baruchiro, line 14: I don't know why only for me the `page.$(selector).click()` don't work, but the `eval` way did.

(I think I was the one that changed this in the `link` click)

esakal: This method is used in other companies so to change it we now need to test them all. 
Maybe it is better that you will copy this method into `mizrahi` scraper and use it locally. That way this PR will not affect other scrapers beside Mizrahi and we will be able to merge it. wdyt?

baruchiro:
@esakal I understand, but it is my old question- how we can test the library every time?

You actually said that we can't change the shared code anymore, because it affects the other scrapers.

In this case, I can't implement it in the scraper because this is about the `login` button, and the base class is clicking it.

I can guess that the `eval` is always better than `$` and `click` because it runs on the browser context?

In my case, the `$.click` way is touching the button but it not clicked (you just see the button color changing like a click effect, but no action).

esakal:
@baruchiro I agree that we currently don't have a way to run regression tests because we rely on private credentials. I don't see a good way to perform regression tests in our use-case, if you have an idea let us know.

Regarding this PR, we can obviously ask people of the affected companies to test this PR and once most of the companies were tested it should be enough. I think you and I already cover most of the companies. 

But since we already have the Typescript branch and we plan to merge it soon. If you approve I can merge this PR into the Typescript one, where we already testing the companies.

wdyt?

baruchiro:
I can redo these changes to the `typescript` branch, but we need all testers to test again?

esakal:
fixed in #500

PR #500, opened 2020-08-02
Fix Mizrahi for TypeScript

baruchiro:
Duplicate of #490 for testing with contributors

### DIFF
--- a/src/scrapers/base-scraper-with-browser.ts
+++ b/src/scrapers/base-scraper-with-browser.ts
@@ -36,7 +36,7 @@ export type PossibleLoginResults = {
 
 export interface LoginOptions {
   loginUrl: string;
-  checkReadiness?: () => Promise<void>;
+  checkReadiness?: () => Promise<unknown>;
   fields: {selector: string, value: string}[];
   submitButtonSelector: string;
   preAction?: () => Promise<void>;
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -5,8 +5,7 @@ import {
 } from '../constants';
 import { BaseScraperWithBrowser, LoginResults, PossibleLoginResults } from './base-scraper-with-browser';
 import { fetchPostWithinPage } from '../helpers/fetch';
-import { waitForNavigation } from '../helpers/navigation';
-import { pageEvalAll } from '../helpers/elements-interactions';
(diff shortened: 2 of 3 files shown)


src/scrapers/base-scraper-with-browser.ts
```diff
@@ -36,7 +36,7 @@ export type PossibleLoginResults = {
 
 export interface LoginOptions {
   loginUrl: string;
-  checkReadiness?: () => Promise<void>;
+  checkReadiness?: () => Promise<unknown>;
```
baruchiro, line 39: I think if you force me to return a `void`, I can't use a **one-line arrow function** because it returning the expression return type.

src/scrapers/mizrahi.ts
```diff
@@ -124,14 +129,22 @@ async function extractPendingTransactions(page: Page): Promise<Transaction[]> {
...
+  await Promise.race([
+    waitUntilElementFound(page, afterLoginSelector),
+    waitUntilElementFound(page, invalidPasswordSelector),
+  ]);
+}
+
 class MizrahiScraper extends BaseScraperWithBrowser {
   getLoginOptions(credentials: ScraperCredentials) {
     return {
-      loginUrl: `${LOGIN_URL}`,
```
baruchiro, line 130: Why was it like that?
(you don't need to answer, just note if I'm missing something.)

src/scrapers/mizrahi.ts
```diff
@@ -124,14 +129,22 @@ async function extractPendingTransactions(page: Page): Promise<Transaction[]> {
...
+      loginUrl: LOGIN_URL,
       fields: createLoginFields(credentials),
-      submitButtonSelector: '#ctl00_PlaceHolderLogin_ctl00_Enter',
-      postAction: async () => waitForNavigation(this.page, { waitUntil: 'networkidle0' }),
-      possibleResults: getPossibleLoginResults(),
+      submitButtonSelector,
+      checkReadiness: async () => this.page.waitForSelector(loginSpinnerSelector, { hidden: true }),
```
baruchiro, line 145: This is about the comment above. If `checkReadiness` is `Promise<void>`, this arrow function not returning `void`

esakal:
scraped successfully Leumi, Isracard, max, visa cal

PR #509, opened 2020-08-16
Adding Identifier to visa-cal

omercnet:
Adding the Id property from VisaCal API to the scraped transactions

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -76,6 +76,7 @@ interface CardByAccountResponse {
 }
 
 interface ScrapedTransaction {
+  Id: string;
   TransType: string;
   Date: string;
   DebitDate: string;
@@ -180,6 +181,7 @@ function convertTransactions(txns: ScrapedTransaction[]): Transaction[] {
   return txns.map((txn) => {
     return {
       type: convertTransactionType(txn.TransType),
+      identifier: parseInt(txn.Id, 10),
       date: moment(txn.Date, DATE_FORMAT).toISOString(),
       processedDate: moment(txn.DebitDate, DATE_FORMAT).toISOString(),
       originalAmount: -txn.Amount.Value,


omercnet:
@esakal @maorsom could you take a look?

eshaham (approved):
Nice 💪 
thanks @omercnet 🙏

omercnet:
bekef :)

would appreciate a patch release so I can use the official library in my project

omercnet:
actually please hold, another push to max on the way

esakal:
@omercnet Hi!
Since @eshaham already merged this PR, can you create a new one based on the latest master that holds your previous changes?

esakal:
> would appreciate a patch release so I can use the official library in my project

@omercnet just to be sure you are familiar with this option. we release automatically a version when pushing to master and deploy it to NPM with tag `next`.

```
```

omercnet:
> @omercnet Hi!
> Since @eshaham already merged this PR, can you create a new one based on the latest master that holds your previous changes?

yes of course, I meant hold off before the patch release :)

sagiba:
Note that the Id attribute by itself is not unique for installment transactions on visa cal.
I've been using the following combo successfully for years now:
```javascript
txn.identifier = `${raw.Id}:${raw.Numerator}`;
```

esakal:
@sagiba Hi
the identifier represents אסמכתא which isn't unique in most companies. For example in bank Leumi you will get the same identifier for every transaction that transfer money to your credit card of leumi card.

what you did will create a unique identifier but is not אסמכתא.

sagiba:
@esakal interesting, what is the אסמכתא alone used for though, if not for uniquely identifying a transaction?

I think it will be super useful for the library to also return a unique ID for each transaction (I already depend on this in my project, to sync with an internal database), but this is discussion is probably worthy of a separate issue.

PR #511, opened 2020-08-18
Added accumulating basket type to max

sushi30:
I hit this transaction type while scraping max website. It is a transaction for loading funds to a gift card (such as [this](https://www.max.co.il/he-il/cards/pages/gcexecutive.aspx/))

image


eshaham (approved):
thanks for the contribution @sushi30 🙏

PR #517, opened 2020-09-09
Make prepareBrowser and preparePage optional in ScraperOptions

brafdlog:
Currently the interface of `ScaperOptions` has `prepareBrowser` and `prepagePage` as required options.

These don't seem like options that should be required because not all use cases need these. And the code itself checks if they exist or not so this will not break the existing code.


esakal (approved):
@baruchiro thanks :)

baruchiro:
> @baruchiro thanks :)

@brafdlog thanks :)

@esakal note :-D

PR #523, opened 2020-10-08
Fix import puppeteer in core library

baruchiro:
I don't know how the `prepare:core` is actually work, but when I'm trying to use `v1.0.0`, I'm getting an error and the problem is in `mizrahi.d.ts` with the line `checkReadiness: () => Promise<import("puppeteer").ElementHandle<Element>>;`.

It shouldn't use `puppeteer` in the `core` library.

So I changed this function to work as the other `elementInteractions` and to return `Promise<void>` so no `puppeteer` in the return type.

### DIFF
--- a/src/helpers/elements-interactions.ts
+++ b/src/helpers/elements-interactions.ts
@@ -5,6 +5,10 @@ async function waitUntilElementFound(page: Page, elementSelector: string,
   await page.waitForSelector(elementSelector, { visible: onlyVisible, timeout });
 }
 
+async function waitUntilElementDisappear(page: Page, elementSelector: string, timeout = 30000) {
+  await page.waitForSelector(elementSelector, { hidden: true, timeout });
+}
+
 async function fillInput(page: Page, inputSelector: string, inputValue: string): Promise<void> {
   await page.$eval(inputSelector, (input: Element) => {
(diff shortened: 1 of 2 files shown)


esakal (approved):

esakal:
Nice workaround. You are right that puppeteer shouldn't be part of core. 

I reviewed the converter we have to create the core version it the bug is not there. 

Mizrahi is returning `this.page.waitForSelector` for `checkReadiness` > `waitForSelector` it returns `ElementHandle` which is not re-exported by the `@types/puppeteer-core` and instead returns that type from `@types/puppeteer` which is why you are getting this problem.

You could instead do the following to solve it:
```
checkReadiness: async () => {
        await this.page.waitForSelector(loginSpinnerSelector, { hidden: true });
      },
```
which will result with what we wished for anyhow `Promise<void>`.  I will fix the declaration of `checkReadiness` to expect for void just to prevent similar cases.

esakal:
@eshaham please release a new official version when possible.

@baruchiro In the meanwhile feel free to pull the next tag from npm:

```
```

esakal:
@baruchiro hold on with the next version. the automatic process is broken, you can track its status here #525

PR #524, opened 2020-10-10
change checkreadiness delcaration

esakal:
following #523 make sure checkreadiness returns Promise<void> to prevent similar issues due to bug in `@types/puppeteer-core`


esakal:
@baruchiro  fyi

eshaham (approved):

PR #535, opened 2020-11-13
revert prerelease.yml file

### LINKED ISSUE #525: Next version automatic deployment is broken and needs new NPM credentials
@eshaham it seems that something happened with the authentication token used to deploy automatically the next version to npm. see the [failure results] (https://github.com/eshaham/israeli-bank-scrapers/runs/1235168325?check_suite_focus=true)

Can you please update the environment authentication token to resolve it? 

you can then click on the re-run jobs to verify that it works.
image

image

esakal:
@eshaham I think it would be better to revert to previous release process. I cannot actually test it before merging it to master but we know that it worked so.... 

wdyt?

closes #525

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -8,21 +8,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
-        with:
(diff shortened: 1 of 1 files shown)


eshaham (approved):

eshaham:
I think something was broken earlier if I remember correctly, but we can try to fix it differently

Issue #550, opened 2021-02-02
Not all tansaction get on Visacal


moshe212:
I do not get my transaction from 2021 at all.
I pay on 10.01.2021 and i do not see this at all..

baruchiro:
More than that, I see the `max.ts` and `visa-cal.ts` shared the code related to `startMoment`, so you need to find where it different.

I guess it something related to the scraping itself.

esakal:
@moshe212  does it still happen to you? if so, can you write here the version you are actually using (either from package.json or package-lock.json) 

also what do you get? empty array or an error?

esakal:
@moshe212 let me know if it still happens. If it does, share more logs / information so we will be able to try and assist

PR #551, opened 2021-02-02
Fix babel and jest

baruchiro:
I tried to add an `expect` with nullish operator, as you can see [here](https://github.com/eshaham/israeli-bank-scrapers/issues/550#issuecomment-771755345), but `jest` throws compilation error because of that.

Then I added the suggestion from https://github.com/kulshekhar/ts-jest/issues/1283#issuecomment-552147794, and it works.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,5 +1,3 @@
-const packageJson = require("./package.json");
-
 const presets = [
   [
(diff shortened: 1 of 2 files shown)


esakal (approved):

esakal:
Thanks @baruchiro!

PR #554, opened 2021-02-24
Patch 1

yacovmalen:


esakal:
@yacovmalen did you create this PR by mistake or on purpose? 
If on purpose, this change breaks the public API something that we try to avoid. regarding the `unknown` instead of `error`. this is something that we can consider by extending the options object with `allowUnknownTransactions: boolean` flag. 

Please reply if it was create on purpose, if so please update the title and add description with relevant information

Issue #397, opened 2020-03-28
Isracard shotef +60


gczobel:
Hi,
I'm having an issue with Business Isracard CC.
These card gives +60 days from the day the charge is created. For example, charges from 01/March are part of the May statement.

Now, in `getAllMonthMoments` in `dates.js`
https://github.com/eshaham/israeli-bank-scrapers/blob/ad5dda57bddbdcfc53aa7db3f92dfec475f3e647/src/scrapers/base-isracard-amex.js#L160

```
async function fetchAllTransactions(page, options, startMoment) {
  const allMonths = (0, _dates.default)(startMoment, true);
```

The `allMonths` are calculated from the startMoment up to the current month + 1. So there is no way to capture the May statement (where the March transactions are saved).

Any idea how to solve this?

esakal:
hi
are those transactions visible in their site? can you provide a snapshot and blur your sensitive information? 

If it does appear there and it is only a start time it should be possible to allow custom end time

gczobel:
Yes, is visible in the site. You only need to change the date to May.
See the screenshots.

![dateselection](https://user-images.githubusercontent.com/754466/77819964-e930b700-70ef-11ea-90f0-505bc1d0b1fd.png)
![CC_Shotef60](https://user-images.githubusercontent.com/754466/77819965-ec2ba780-70ef-11ea-8380-7064d4d930db.png)

gczobel:
@esakal No workaround for this?

esakal:
Hi,
I can mark this request as an enhancement and we can add it to the queue. It might take a while thou because we have some infrastructure improvements we are planning and has higher priority.

In the meanwhile you can easily fork this library and modify the code, it is quite easy as you don't really need to consider breaking api changes and other things that we need to think about before adding it to the official version:

1. Fork this repository
2. in [dates.js line 9](https://github.com/eshaham/israeli-bank-scrapers/blob/ae252e2991fddb174044f9dab00af482b5871904/src/helpers/dates.ts#L9) change number `1` to `2`.
3. consider testing the library as mentioned here
4. `npm run build` the library. 
5. `npm pack` to create a `tgz` file of your own version.
6 copy the tgz file into your project into `libs/file-name.tgz`
7. run `npm i file:libs/file-name.tgz` (obviously replace `file-name` here with the actual file name)

Let me know if it helped.
Eran

gczobel:
The workaround is ok...
But please, mark the issue as a enhancement.

[...]

PR #560, opened 2021-03-24
fix: add endDate to base-scrapper

gczobel:
This is a fix (enhancement) for #397 adding endDate to the scrapper options.
I tested it using Isracard and Max cards on the real websites


esakal:
Hi @gczobel  thank you for your contribution, I have some concerns with this PR:
1. the important one is that although this is a general property, it is supported only in 3 scrapers, for the rest it is ignored.
1. It is not backward compatible. I believe it is by mistake but you are by default fetching only one month if I understood the code correctly
2. if we will continue with this change it should be reflected in the documentation that there is such an option.

@eshaham what do you think?

PR #565, opened 2021-04-18
Fix Bank Leumi scraper

### LINKED ISSUE #564: Bank Leumi - changed the transaction page and it seems to be broken..
Its look like there is a version change (from ASP.net to angular)
The issuse starts with leumi.ts file row 246: "const accounts = await dropdownElements(page, 'select#ddlAccounts_m_ddl');" account exist but script not find them,

esakal:
Update Leumi scraper to adjust to the new site that is using Angular now.

There are three things to know about:
1. this scraper supports single account, as I don't have multiple accounts. Whoever in the future that will want to support multiple accounts would need to extend the scraper to support multiple accounts - **This is not a blocker, I believe once we confirm the rest we will be able to merge it as-is**
2. I didn't check yet the pending transactions as I don't have them, I will verify it later in the weekend.
3. I don't remember if the account number included the snif or not, I just picked the one showed in the site.

@ifeins please pull this PR and test my scraper. Let me know how it went. 

closes #564

### DIFF
--- a/src/scrapers/leumi.ts
+++ b/src/scrapers/leumi.ts
@@ -126,15 +126,15 @@ function hangProcess(timeout: number) {
 async function fetchTransactions(page: Page, startDate: Moment): Promise<TransactionsAccount[]> {
   // TODO should adjust logic to support multiple accounts (I don't have such an account)
 
-  // DEVELOPER NOTICE the account number received from the server is being altered at runtime for some accounts
-  // after 1-2 seconds so we need to hang the process for a short while.
+  // DEVELOPER NOTICE the account number received from the server is being altered at
+  // runtime for some accounts after 1-2 seconds so we need to hang the process for a short while.
   await hangProcess(4000);
 
   const accountSpanText = await page.$eval('app-masked-number-combo span.display-number-li', (span: any) => {
     return span.textContent;
   });
 
-  // due to a bug, the altered value might include undesired signs like & that are removed here
+  // due to a bug, the altered value might include undesired signs like & that should be removed
   const accountNumberMatches = accountSpanText.match(/\d+-\d+(\/)?\d+/);
   const account = accountNumberMatches ? accountNumberMatches[0] : null;
 
--- a/src/scrapers/leumi.ts
+++ b/src/scrapers/leumi.ts
@@ -2,15 +2,11 @@ import moment, { Moment } from 'moment';
 import { Page } from 'puppeteer';
 import { BaseScraperWithBrowser, LoginResults, LoginOptions } from './base-scraper-with-browser';
 import {
-  dropdownSelect,
-  dropdownElements,
   fillInput,
   clickButton,
   waitUntilElementFound,
   pageEvalAll,
-  elementPresentOnPage,
 } from '../helpers/elements-interactions';
-import { waitForNavigation } from '../helpers/navigation';
 import { SHEKEL_CURRENCY } from '../constants';
 import {
   TransactionsAccount, Transaction, TransactionStatuses, TransactionTypes,
@@ -18,13 +14,12 @@ import {
 import { ScaperScrapingResult, ScraperCredentials } from './base-scraper';
 
 const BASE_URL = 'https://hb2.bankleumi.co.il';
-const DATE_FORMAT = 'DD/MM/YY';
-const NO_TRANSACTION_IN_DATE_RANGE_TEXT = 'לא קיימות תנועות מתאימות על פי הסינון שהוגדר';
+const TRANSACTIONS_URL = `${BASE_URL}/eBanking/SO/SPA.aspx#/ts/BusinessAccountTrx?WidgetPar=1`;
+const FILTERED_TRANSACTIONS_URL = `${BASE_URL}/ChannelWCF/Broker.svc/ProcessRequest?moduleName=UC_SO_27_GetBusinessAccountTrx`;
+
+const DATE_FORMAT = 'DD.MM.YY';
 const ACCOUNT_BLOCKED_MSG = 'המנוי חסום';
 
-function getTransactionsUrl() {
-  return `${BASE_URL}/ebanking/Accounts/ExtendedActivity.aspx?WidgetPar=1#/`;
-}
 
 function getPossibleLoginResults() {
   const urls: LoginOptions['possibleResults'] = {};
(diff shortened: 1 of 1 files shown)


ifeins:
@esakal Tested it using `israeli-ynab-updater` (using `npm link israeli-bank-scrapers`) and it worked!
Thanks for the fix 👍 

```
scraping Bank Leumi
Bank Leumi: START_SCRAPING
Bank Leumi: INITIALIZING
Bank Leumi: LOGGING_IN
Bank Leumi: LOGIN_SUCCESS
Bank Leumi: TERMINATING
Bank Leumi: END_SCRAPING
success: true
Bank Leumi: scraped 23 transactions from account ...
exporting 23 transactions for account ...
1 csv files saved under ...
➜  israeli-ynab-updater git:(upgrade_node_version)
```

eshaham:
@esakal since they changed to Angular, don't they use a REST call to get the transactions from within their app?
if they do, we should switch to that, like in DiscountScraper, it will be much more reliable

esakal:
@eshaham they are doing a mix, they don't have a pure API, you can check the network calls, they add some information that is being calculated and probably vary between users. It usually like that for sites that are rewritten in steps and are not a full refactor with different code-base.

So I cannot predict the API and thus cannot do the same as Discount. But the good news are that this implementation is actually much more reliable then what we had in Bank Leumi because one I simulate the right call to the server, they return a single response without pagination which is structured. As you saw in the provided changes, I'm not scrapping HTML table anymore, instead I'm mapping directly the JSON they provide into our structure.

erikash:
@esakal 
Hi,
I've tested this with my Leumi user and unfortunately your logic doesn't scrape *all the accounts* for a given user, it scrapes just the default one...

Please advise...

esakal:
@erikash I know; this is an unfortunate limitation that I mentioned in the PR description. Since I don't have multiple accounts, I can fix the scraper to support a single account but not multiple ones.

The good news is that I lay out the ground to support multiple accounts. Someone with basic node.js knowledge should be able to adjust the new scraper and support multiple quickly.

As I see it, there are several ways to handle it. The optimal one is if you have basic knowledge in node.js, and then I can guide you where to add that support in the scraper. Otherwise, we should wait for other people who are using this library to notice the issue, and I believe one should be able to do it. 

If it will not happen and it will continue to bug you, we can schedule a short live session where I will remote into your machine and will add that support; remoting into your device means that I will be exposed to your data on Bank Leumi so you should take it into account. Suppose for any reason, like lack of IDE, it is not possible, and you really trust people in general. In that case, you can provide me the credentials during a live session and watch me fix it on my machine; and after I adjust the scraper, you can change your password, but again you should feel comfortable with that option.

So as I see it at the moment, the best way to handle it is either you add it with my guidance, or we wait for someone else who has the same issue and can adjust the scraper as I mentioned above.

esakal:
@eshaham good news! pending transactions works, I just confirmed it. As I see it, we can proceed with the merge to master.
Regarding support for multiple accounts, there no need to wait for it, it is better to support the default and once someone will really need it he will be able to create a PR with some adjustments (see comment above to @erikash

ifeins:
@esakal There is a small issue with the output of the bank leumi CSV file, not sure if's related to `israeli-bank-scrapers` or to `israeli-ynab-updater` but since it previously worked ok then there is cause to suspect it might be related to this change.

This is the output of files I get after running the ynab-updater:

```
➜  23-04-2021_17-36-33 ll
Bank Leumi (xxx-xxxxx
Isracard (xxxx).csv
Isracard (xxxx).csv
Visa Cal (xxxx).csv
➜  23-04-2021_17-36-33 ll Bank\ Leumi\ \(xxx-xxxxx
33).csv
```

It seems that the Bank Leumi filename gets cut and so instead of the output to be in a single file `Bank Leumi (xxx-xxxxx33)` It gets split into a folder `Bank Leumi (xxx-xxxxx` and file `33).csv`.

The `33` is probably the "numerical code" for balance accounts (עובר ושב) (at least in Bank Leumi the balance account always end in `33`).

erikash:
I saw it too..
It’s a breaking change @esakal
The account was formatted as xxxxxx_xx and now it’s xxxxxx/xx (which
probably explains the inner folder created)

esakal:
Hi @ifeins you are right, the issue you have in the other library happens due to the fix of the scraper. the account number of Bank Leumi includes the sign `/`, for example `xxx-xxxxxx/xx`, and in the previous scraper it somehow altered the account number to not include it.

I'm not sure what was the previous format, and also I think scrapping the original value is what we should do in that library. So the fix should be in `israeli-ynab-updater`. I'm not using that library but I can guide you what to fix so you will be able to create a PR, it is easy. You should replace in file [generate-reports.js line 72](https://github.com/eshaham/israeli-ynab-updater/blob/75335bc919b086deedc75fc8aab15e7410344ceb/src/scrape/generate-reports.js#L72) the content to:

```
const formattedAccountNumber = accountNumber.replace('/', '_');
await writeFile(`${saveLocation}/${scraperName} (${formattedAccountNumber}).csv`, csv);
```

As we replace the sign `/` with `_` the issue you described will not happen anymore. You can change it to other sign like `-` or even to nothing ``, up to you to decide.

I hope this help
Eran

erikash:
+1
For keeping the new / behavior

I do think this is a breaking change though..
But probably not a major one and could be considered a bug fix.

esakal:
@eshaham what do you say? can we merge it? the current scraper is broken, no-one can scrape Bank Leumi and also the library return a valid response with 0 accounts which is unfortunate. This PR was tested already by myself and three more people that responded in the issue so I believe we can continue and release a new version with that.

To save time, please merge on your own and release a version at the same time, wdyt?

Thanks
Eran

esakal:
@erikash @ifeins I see that in union bank, masaad and bein leumi we do replace `/` with `_` for the account number. Following @erikash comment and as fixing it in Bank Leumi doesn't resolve it cross scrapers, I also replaced the symbol `/` with `_` in Bank Leumi - xxx-xxxxxx_xx

@eshaham fyi

erikash:
Ok, thanks.

src/scrapers/leumi.ts
```diff
@@ -51,206 +46,105 @@ function createLoginFields(credentials: ScraperCredentials) {
...
+      resolve();
+    }, timeout);
+  });
+}
+async function fetchTransactions(page: Page, startDate: Moment): Promise<TransactionsAccount[]> {
+  // TODO should adjust logic to support multiple accounts (I don't have such an account)
+
+  // DEVELOPER NOTICE the account number received from the server is being altered at
+  // runtime for some accounts after 1-2 seconds so we need to hang the process for a short while.
+  await hangProcess(4000);
```
eshaham, line 131: I don't like this - what's the value of the field before it gets updated?
we can easily do a validation that the account number is in the right format, and wait until that's the case.
we should also use the `waitUntil` helper function for this sleep.

eshaham (changes requested):
@esakal added a small comment

src/scrapers/leumi.ts
esakal: we cannot because this is a bug on their side that is being fixed and it can vary between users. for example I'm getting xxx-xxxxxx/46 which is then changed to xxx-xxxxxx/26 due to hard-coded logic that was added, I believe as a workaround to the wrong value returned from the server. @ifeins gets `/33`, there are multiple types and we cannot predict when they run their workaround.

I believe in the future we will be able to remove it, but at the moment, as much as I hate adding timeouts, this time I believe it is the best solution we have.

eshaham: @esakal got it - can we then just switch to using `waitUntil`?

erikash:
Can we iterate all available accounts from a dropdown list ?

esakal:
> Can we iterate all available accounts from a dropdown list ?

@erikash yes we can, but see my comment [here](https://github.com/eshaham/israeli-bank-scrapers/pull/565#issuecomment-823797161) about it

erikash:
I try to look into it this week..
I asked in the context of account numbers changing dynamically, can we use
another source for the account numbers (such as a drop-down list)?

esakal:
> I asked in the context of account numbers changing dynamically, can we use...

Yes for sure, you can see what I did previously in the leumi, there we also had dropdown. Your code change should be in this function

https://github.com/eshaham/israeli-bank-scrapers/blob/6d563c7021891b731e6ee61828419b2c32d090b3/src/scrapers/leumi.ts#L127

I did some preparation there. Because I have only one account I don't have that dropdown. you should keep it in mind and fallback to what I did if you don't find the dropdown element to support both multiple and single account.

Feel free to consult and we can always do a short live session using discord or something similar (subject to what I said that it means that I will be exposed to your financial information)

izikf89:
Hi @esakal,
I've tested and its work to me for default account, i changed code for its supports multi accounts, open a pull request?

esakal:
@izikf89  that is great! open a PR against branch `leumi-to-angular`

Thanks
Eran

baruchiro:
@esakal @eshaham , when you're going to merge it?
I think we need a short cycle for scraping fixes. For now, I'm not getting my updated transactions data for two weeks ago.

izikf89:
> @izikf89 that is great! open a PR against branch `leumi-to-angular`
> 
> Thanks
> Eran

I do not have permission to open brunch

esakal:
@izikf89 you should do it from a fork, see [documentation](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork)

@baruchiro - @eshaham and I will figure a way to make this process shorter.

eshaham (approved):

PR #566, opened 2021-04-24
fixed union bank scraper

orentet:
also, please add me to the contributors of union bank scraper.

### DIFF
--- a/src/scrapers/union-bank.ts
+++ b/src/scrapers/union-bank.ts
@@ -274,10 +274,7 @@ async function fetchAccounts(page: Page, startDate: Moment) {
 }
 
 async function waitForPostLogin(page: Page) {
-  return Promise.race([
-    waitUntilElementFound(page, '#signoff', true),
-    waitUntilElementFound(page, '#restore', true),
-  ]);
+  return waitUntilElementFound(page, '.signoff_button', false);
 }
 
 class UnionBankScraper extends BaseScraperWithBrowser {
--- a/src/scrapers/union-bank.ts
+++ b/src/scrapers/union-bank.ts
@@ -210,7 +210,7 @@ async function searchByDates(page: Page, startDate: Moment) {
 
 async function getAccountNumber(page: Page) {
   const selectedSnifAccount = await page.$eval('#ddlAccounts_m_ddl option[selected="selected"]', (option) => {
-    return (option as HTMLElement).innerText;
+    return (option as HTMLElement).innerText.replace(/[^ -~]+/g, '');
   });
 
   return selectedSnifAccount.replace('/', '_');
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Any kind of help is welcome, even if you just discover an issue and don't have t
 While there's no specific template for creating a new issue, please take the time to create a clear description so that it is easy to understand the problem.
 
 ## Testing the scrapers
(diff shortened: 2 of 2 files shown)


src/scrapers/union-bank.ts
```diff
@@ -210,7 +210,7 @@ async function searchByDates(page: Page, startDate: Moment) {
 
 async function getAccountNumber(page: Page) {
   const selectedSnifAccount = await page.$eval('#ddlAccounts_m_ddl option[selected="selected"]', (option) => {
-    return (option as HTMLElement).innerText;
+    return (option as HTMLElement).innerText.replace(/[^ -~]+/g, '');
```
eshaham, line 213: @orentet why is this needed?
missing an explanation for this change in your PR

src/scrapers/union-bank.ts
```diff
@@ -274,10 +274,7 @@ async function fetchAccounts(page: Page, startDate: Moment) {
 }
 
 async function waitForPostLogin(page: Page) {
-  return Promise.race([
-    waitUntilElementFound(page, '#signoff', true),
-    waitUntilElementFound(page, '#restore', true),
-  ]);
+  return waitUntilElementFound(page, '.signoff_button', false);
```
eshaham, line 277: @orentet why are we changing this logic? doesn't it work as it is?

eshaham (changes requested):
@orentet thanks for your contribution
can you please add a description to your PR to explain your changes?

src/scrapers/union-bank.ts
orentet: no. the UI was probably changed.

src/scrapers/union-bank.ts
orentet: when selecting the account, it contains non-printable chars on both sides, which this code deletes.

orentet:
any change?

esakal:
@orentet Hi,
the PR is missing a description and it is not clear from the code change why you introduce those changes. Please add description explaining your changes. Thanks

orentet:
Hello Eran,
Please look at my responses for the code changes.
The reason for the changes is because union bank is currently not working
because of UI changes, and my code fixes it.

src/scrapers/union-bank.ts
esakal: maybe I'm wrong but when using `[^ -~]+` pattern (starting with ^ symbol in the brackets) you are basically replacing everything that is not one of those characters, so you keep those signs and not the actual account value. where am I mistaken?

regarding that non-printable sign, I tried your regexp in two online editors and that sign broke them. can you instead use the hex representation of that sign, or maybe use allowed list instead. for example `/[\d-]+/g`

src/scrapers/union-bank.ts
```diff
@@ -274,10 +274,7 @@ async function fetchAccounts(page: Page, startDate: Moment) {
 }
 
 async function waitForPostLogin(page: Page) {
-  return Promise.race([
-    waitUntilElementFound(page, '#signoff', true),
-    waitUntilElementFound(page, '#restore', true),
```
esakal, line 279: why did you remove the #restore part? this is basically the view that is shown once in a while to replace a password? I don't have union so I cannot verify it but based on other scrapers usually here the scraper wait for either a valid login or a password expired status hence the promise.race

consider keeping the restore part

esakal:
@orentet thanks for the clarification. The Github code review dashboard is extremely annoying and misleading. In my workplace we use [https://reviewable.io/](https://reviewable.io/) which is must superior from the Github one.  Please don't resolve the comments, just add your insights on them, the resolution should be done only once we all agree.

I added my insights about your changes, let me know what you thinkg

PR #575, opened 2021-05-10
automate the deployment process to NPM using semantic releases

esakal:
@eshaham important - don't squash this one specifically. **use merge please**


.github/workflows/nodeCI.yml
```diff
@@ -4,14 +4,12 @@ on: [push, pull_request]
 
 jobs:
   build:
-
+    if: github.ref != 'refs/heads/master'
```
eshaham, line 7: any reason not to run this during merge to master?

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+      - name: Release core to NPM
+        if: steps.semantic.outputs.new_release_published == 'true'
+        run: |
+          npm run prepare:core
+          node utils/pre-publish.js --version ${{ steps.semantic.outputs.new_release_version }}
+      - run: npm publish --access public --ignore-scripts
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
eshaham, line 61: add line break 🙏

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+                prerelease: true
+              },
+              {
+                name: 'alpha',
+                prerelease: true
+              }
+            ]
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Release default to NPM
```
eshaham, line 47: this is not the release script, it is preparing the package.json for release

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        if: steps.semantic.outputs.new_release_published == 'true'
+        run: |
+          node utils/pre-publish.js --version ${{ steps.semantic.outputs.new_release_version }}
+      - run: npm publish --access public --ignore-scripts
```
eshaham, line 51: would be great to name this step

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+      - name: Release core to NPM
```
eshaham, line 54: this is not the release script, it is preparing the package.json for release

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        if: steps.semantic.outputs.new_release_published == 'true'
+        run: |
+          npm run prepare:core
+          node utils/pre-publish.js --version ${{ steps.semantic.outputs.new_release_version }}
+      - run: npm publish --access public --ignore-scripts
```
eshaham, line 59: would be great to name this step

package.json
```diff
@@ -21,8 +22,8 @@
     "build:js": "babel src --out-dir lib --extensions \".ts\" --source-maps inline --verbose",
     "prebuild": "node utils/update-puppeteer-config.js",
     "postbuild": "rimraf lib/tests",
-    "prepare:core": "node utils/prepare-israeli-bank-scrapers-core.js && npm i --package-lock-only && npm ci && npm run lint -- --fix && npm run build",
-    "prepare:default": "npm ci && npm run build",
+    "prepare:core": "git reset --hard && node utils/prepare-israeli-bank-scrapers-core.js && npm i --package-lock-only && npm ci && npm run lint -- --fix && npm run build",
+    "prepare:default": "git reset --hard && npm ci && npm run build",
     "reset": "git reset --hard && npm ci",
     "publish:core": "npm run prepare:core && npm publish --access public && npm run reset",
     "publish:default": "npm run prepare:default && npm test && npm publish --access public && npm run reset",
```
eshaham, line 29: these shouldn't work anymore, right?
I mean, if we want to keep them for manual publish, perhaps we should fix them as well?

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+                name: 'alpha',
+                prerelease: true
+              }
+            ]
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Release default to NPM
+        if: steps.semantic.outputs.new_release_published == 'true'
+        run: |
+          node utils/pre-publish.js --version ${{ steps.semantic.outputs.new_release_version }}
```
eshaham, line 50: would be great to add this to an npm script we call here 🙏

eshaham (changes requested):

.github/workflows/nodeCI.yml
esakal: yes, if we merged to master it means that the PR we just merged already ran it. Also, even if it will fail the version will be created already. and during creation of a version we already run linter. Actually I don't think that workflow is useful as we don't have concrete logic, running it as part of PRs seems good enough.

.github/workflows/release.yml
esakal: I don't think workflows yml based usually have line breaks

.github/workflows/release.yml
esakal: done

.github/workflows/release.yml
esakal: done

.github/workflows/release.yml
esakal: done

package.json
esakal: removed

.github/workflows/release.yml
esakal: it is not intuitive as we pass here version as args, I prefer leaving it like that

.github/workflows/nodeCI.yml
eshaham: 👍

.github/workflows/release.yml
eshaham: 👍

.github/workflows/release.yml
esakal: I though you asked for breaklines between lines not at the end. 
If you are referring to the end of the file like the tooltip mention, I'm not sure why it writes it because i do have a line break, see snapshot

image

eshaham: weird, probably one of the newer commits fixed it 👌

eshaham (approved):

esakal:
@eshaham any idea how to force the lint status to run? It happened due ton outrage Github action had a few days ago but it prevents me from merging. maybe enforce a re-run of the linter or bypass this check and merge it?

PR #582, opened 2021-06-15
Removed unnecessary backticks  so code will be node-flavored

itainathaniel:
This is very simple PR, just removing 3 backticks form the README file so the code block will be shown as node.


esakal (approved):
Hi @itainathaniel thanks for fixing this one

esakal:
@itainathaniel Thanks! indeed it is small, but Github has a ghost check `lint` that prevent us to merge it. We will figure it up soon

PR #593, opened 2021-07-01
Leumi support CHANGE_PASSWORD

baruchiro:


esakal (approved):
Thanks @baruchiro 👍

esakal:
@baruchiro I approved the PR but I"m having some technical difficulties to merge it. We have an orphaned status check named `lint` which prevent us from merging unless you bypass it with super user rights (which only @eshaham have them). Anyhow using the super user rights is not the best way to handle it. and we should resolve it. I opened #594 which should restore the missing status check.

@eshaham decide if you want to bypass that status check one more time here or merge the PR I created so @baruchiro will be able to merge into his PR and we will hopefully be able to merge it as expected.

baruchiro:
@eshaham Added the same fix for Mizrahi

eshaham (approved):

PR #594, opened 2021-07-01
chore: restore lint status check which is blocking merges

esakal:
@eshaham at the moment due to the orphaned status check `lint` that blocks all PRs I cannot merge approved ones like #593 

As I didn't find a way to get rid of this status check, I though maybe it is easier to just add it again so it will hopefully resolve the blocker we currently have on any PRs.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,18 @@
+name: lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Run npm install
+        run: npm install
+
+      - name: Run lint
+        run: npm run lint
+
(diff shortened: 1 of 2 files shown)


baruchiro:
In my opinion, we have to run `lint` on PR.

esakal:
@baruchiro we run lint on any PR as part of another workflow https://github.com/eshaham/israeli-bank-scrapers/blob/f083975993052c9d8b21bec464fbebbae300dd07/.github/workflows/nodeCI.yml#L22

There is no point running the lint also separately. But because the technical issue I mentioned, I'm moving the lint test back to a separate workflow.

baruchiro:
OK. It is not a technical issue, I think. It is just @eshaham need to remove the "required" from this status check.

esakal:
@eshaham please review, it blocks all the other PRs.

esakal:
@eshaham  can you help with that one?

eshaham (approved):
@esakal sorry tough couple of weeks, approving and merging

PR #603, opened 2021-09-01
change the way scraper detect installments

esakal:


ifeins:
Don't know if it's related but I've received the following error while scraping Visa Cal:

```
Error: unknown transaction type null
```

erikash:
@ifeins @esakal same for me...

ifeins:
I was able to scan Visa Cal using this branch, however it seems that it duplicated transactions.
Will follow up to see if it happens next time.

esakal:
@ifeins I'm not sure how you got duplicated transactions. As I see it they just stopped providing some information like transaction type and memo in that API. The only idea I have is that maybe they showed the transaction once in pending and another time as regular and then it might be shown as duplicated one.

ifeins:
@esakal I scraped again and this time it worked fine (no duplicated transactions).

esakal:
@eshaham can you please review and process this one?

eshaham (approved):
👌

erikash:
@esakal @eshaham 
I've tested the scraper and according to my findings the `normal` transactions scrape correctly but the `installments` transactions *get filtered*.

I tried to debug the code and i do see the installment transactions in `rawTxns` (with `Date` of the installment setup date which could be a few months ago)

I didn't have time to dig in further... but my guess is that the semantics of the `Date` changed as well and therefore the transactions get filtered...

erikash:
I've dug some more into this and found out the the `nodeFetch` result for `Transactions` fetches a field called `Numerator` which maybe could be used for differentiating `installment` transactions from `normal` ones...

For my transactions it was `2` for all הוראות קבע and `8` or `12` for installment transactions (not sure what is the difference between `8` and `12`)
image

I saw in `master` that you use a `switch` on `tnxType` to differentiate and `8` for `tnxType` also means installments but this could be a coincidence.. 

@esakal @eshaham what do you think?

erikash:
Another possible workaround could be inferring that the transaction is an `installment` transaction by comparing the `DebitAmount` and the `Amount`
Combined with `Date` we can calculate:
1. Number of installments =  `Amount` \ `DebitAmount`
2. Payment no - Month different between today and `Date`

It's hacky.. I know.. but WDYS?

esakal:
I think there are too many issues with the changes they introduced in the API we used till now.

I started to create a new scraper using puppeteer, you can track the progress in #609.

erikash:
Thanks Eran! :)

esakal:
@erikash @ifeins sorry for the delay... holidays..... I want to consult with you about how to tackle this one. 

Initially, while working on the scraper, I didn't find any noticeable benefit with the puppeteer. In the current API, we are missing the memo and the type; the transaction type is not exposed as part of the data rendered by the server, so we didn't gain anything useful with the new API.

But we might still need to use the new puppeteer due to the transaction type and refunds.

## Installment transaction
I don't have any installment transactions, so I cannot conclude whether the current API is better than the puppeteer library.
If you can share the JSON structure returned from the current API and a snapshot of the same transaction in the browser, it might help.

Maybe we will be able to use the memo content to guess the type, and once we have it, we can then do a second call to get the needed information by simulating a user that clicks on that transaction. We cannot do that call for every transaction, but I believe people will usually have zero to few installment transactions so that we can make the extra request only for them.

This is relevant only if they appear with the correct date in the browser, and those will not be filtered out by the date range provided.

## Regular transactions
I think all the other data needed can be retrieved from the table without the need to for the additional request
CleanShot 2021-10-01 at 09 58 33

I think we get all this data from the current API besides the memo content, but I might overlook something.

## Refunds
Visa cal is very annoying when it comes to refunds; they show them with the תאריך עסקה and not with the refund date. Thus, you will get them only when you provide a relevant date range.

Still, from my tests, the current API doesn't return them at all.

See this example, I got the refund two weeks after I saw the improper payment, in the browser I get it, but with the תאריך עסקה, still in the current API, I don't get it at all.
CleanShot 2021-10-01 at 10 31 59

---

What do you think?

erikash:
Hi 👋
I lost you for a bit there Eran 😬
Not sure to what APIs are you referring to? Are there 3? Old, New and
Puppeteer?

I would love to help you so feel free to shoot me an email to:
***@***.*** to schedule some time to discuss :)

Erik.

ifeins:
I got the duplicate transactions issue again in YNAB, they were transactions that previously scanned and reconciled and when I scanned again and imported them to YNAB they appeared as duplicates. This didn't happen with the previous scraper. I guess somehow the data format changed and YNAB can no longer correlate and indicate that they are the same transactions.

PR #605, opened 2021-09-08
feat: beinleumi scrapper - balance support

LiranBri:


src/scrapers/base-beinleumi-group.ts
```diff
@@ -273,6 +276,14 @@ async function fetchAccounts(page: Page, startDate: Moment) {
   return accounts;
 }
 
+async function getCurrentBalance(page: Page) {
```
eshaham, line 279: can you move this above the function which calls it?

src/scrapers/base-beinleumi-group.ts
```diff
@@ -273,6 +276,14 @@ async function fetchAccounts(page: Page, startDate: Moment) {
...
+  const balanceStr = await page.$eval(CURRENT_BALANCE, (option) => {
+    return (option as HTMLElement).innerText;
+  });
+  const balance = getAmountData(balanceStr)
+  return balance;
```
eshaham, line 284:
```suggestion
  return getAmountData(balanceStr);
```

eshaham (changes requested):
Hi @LiranBri,
Thanks for contributing.
Added small comments.
Also, notice that your PR doesn't pass the build - can you take a look?

src/scrapers/base-beinleumi-group.ts
LiranBri: I can but In my opinion it's good practice for clean code to put the higher-level function below the lower level, because then you can read the code from top to bottom.
reading first the lower level implementation makes no sense on its own because there is no context. so you always want to read the higher-level code first, and only then dig down to the lower level implementations (the sub-function it triggers)

LiranBri:
> Also, notice that your PR doesn't pass the build - can you take a look?
where can I see that? can you share a link or the error?
I see just list of build pipelines with status they are waiting for status change before they could run (I guess waiting for PR approval of maintainer)

src/scrapers/base-beinleumi-group.ts
eshaham: @LiranBri that's a matter of taste 😉  plus that's how we have the lint rules set up for the entire project at the moment

eshaham:
> where can I see that? can you share a link or the error?
you should see it now - you can also run `npm run lint` locally to see the lint errors

src/scrapers/base-beinleumi-group.ts
LiranBri: IMO its a bad practice, but your project, your call :)
fixed

PR #606, opened 2021-09-11
fix: Workaround for Mizrahi bank API bug which returns transactions before the start date

alonl:
For some reason Mizrahi bank API returns transactions before the requested "from" date (start date). From what I've noticed, usually it is one additional transaction, and it doesn't matter what its date is.

With YNAB exporter in [Caspion](https://github.com/brafdlog/caspion) it does a real mess since it makes those transactions being imported again and again and creates duplicates.

This fix is a workaround for this bug of Mizrahi Bank API which double checks the transaction date is within range.

### DIFF
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -176,7 +176,7 @@ class MizrahiScraper extends BaseScraperWithBrowser {
 
     // workaround for a bug which the bank's API returns transactions before the requested start date
     const startMoment = getStartMoment(this.options.startDate);
-    const oshTxnAfterStartDate = oshTxn.filter(txn => !moment(txn.date).isBefore(startMoment));
+    const oshTxnAfterStartDate = oshTxn.filter(txn => moment(txn.date). isSameOrAfter(startMoment));
 
     await this.navigateTo(PENDING_TRANSACTIONS_PAGE, this.page);
     const pendingTxn = await extractPendingTransactions(this.page);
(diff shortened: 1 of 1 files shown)


src/scrapers/mizrahi.ts
```diff
@@ -172,10 +174,14 @@ class MizrahiScraper extends BaseScraperWithBrowser {
     const relevantRows = response.body.table.rows.filter((row) => row.RecTypeSpecified);
     const oshTxn = convertTransactions(relevantRows);
 
+    // workaround for a bug which the bank's API returns transactions before the requested start date
+    const startMoment = getStartMoment(this.options.startDate);
+    const oshTxnAfterStartDate = oshTxn.filter(txn => !moment(txn.date).isBefore(startMoment));
```
eshaham, line 179:
```suggestion
    const oshTxnAfterStartDate = oshTxn.filter(txn => moment(txn.date). isSameOrAfter(startMoment));
```

eshaham (changes requested):
Thanks for contributing 😄 
Question about the bug with the bank's API - are the transactions returned missed by more than one day?
Meaning, if I specify the start date as 10/7, will it also return transactions from 5/7?

alonl:
> Thanks for contributing 
> Question about the bug with the bank's API - are the transactions returned missed by more than one day?
> Meaning, if I specify the start date as 10/7, will it also return transactions from 5/7?

Thank you! 
Yes. From my observation it seems that it returns one additional transaction, which is the last one that's happened before the start date. It can be one day before or even 2 weeks.

eshaham:
@alonl the lint job is failing - can you run `npm run lint` locally and fix the lint errors?

alonl:
> @alonl the lint job is failing - can you run `npm run lint` locally and fix the lint errors?

Fixed, sorry.

eshaham (approved):
looking good, thanks 👌

PR #615, opened 2021-10-19
Scrape multiple accounts for visaCal

sushi30:
closes https://github.com/eshaham/israeli-bank-scrapers/issues/618

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
   root: true,
   "rules": {
+    "no-nested-ternary": 0,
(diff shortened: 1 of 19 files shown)


esakal (approved):

esakal:
@sushi30 I'm going to merge this one to support multiple cards. I'm going to try a better workaround. I suspect the wait for navigation just doesn't work there as expected. When it fails I see in the snapshot we take that it managed to navigate but for an unknown reason, Puppeteer doesn't detect it. I will let you know if the creative workaround works so you will be able to test it

PR #616, opened 2021-10-20
fix: add debug and resolve visa-cal navigation timeout issue

esakal:
# Motivation
- [x] figure out why in headless mode we get navigation timeout many times for visa-cal
- [x] introduce debug logs into the library and use it with visa-cal  
- [x] add a snapshot on failure capability that is extremely useful when trying to debug issues

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
   root: true,
   "rules": {
+    "import/prefer-default-export": 0,
     "no-nested-ternary": 0,
     "arrow-body-style": 0,
     "no-shadow": 0,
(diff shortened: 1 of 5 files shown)


esakal:
@sushi30 following your comment in #615 I decided to tackle it now and thanks to the logs that I added I figured out that when using headless browser the visa cal page doesn't always react to the filtering submit button as it happens to fast. 

If I'm right you will not have this issue anymore.

Please test is with this codebase. If it still happens run it with `DEBUG=israeli-bank-scrapers*` env set and paste the logs. If it works I will merge this one and #615 right after.

Keep my updated.

@eshaham  @erikash fyi

PR #619, opened 2021-10-30
fix the network and submit race

esakal:
This pull request is yet another attempt to resolve the navigation issue with visa-cal scraper when using `showBrowser` set to `false`.

@sushi30 @RoiSec please try it out - Thanks!

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -12,7 +12,6 @@ import {
   TransactionTypes,
 } from '../transactions';
 import { ScaperOptions, ScaperScrapingResult, ScraperCredentials } from './base-scraper';
-import { waitForNavigationAndDomLoad } from '../helpers/navigation';
 import {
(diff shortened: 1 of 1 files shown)


sushi30:
this works for me in both headless and non-headless mode.

esakal:
@eshaham can you please review this one, it is a small yet important adjustment of the flow

@brafdlog I suggest that once it will be merged you will update caspion app with this version, hopefully it will help your users

eshaham (approved):

erikash:
🥳 
Well done @esakal and @sushi30!

erikash:
@esakal do we still need the `await page.waitFor(1000);`?

esakal:
@erikash I believe we do as they address two different issues if I'm not mistaken. Before we waited for a second, it never worked with a headless browser. While diagnostic the problem I saw in the failed snapshot that we didn't even transmit the request. Once we waited a second it started to work for you and me but failed for others in computers where the navigation started before puppeteer registered to the loading events.

PR #620, opened 2021-10-31
Isracard max include next 3 months

erikash:
Hi,
My payment terms in Isracard for abroad transactions are +1 month (but for some it's +3 months) from the purchase date.
This means that if I bought something in October it will not be processed in November as an ordinary domestic transaction, it will be processed at December. (see screenshot)
image

The current code scrapes transaction to be processed only one month head (November) and therefore misses the two transactions in the screenshot above even though the purchase date is in the past (it's Oct 31 at the time of opening this PR).

I've changed the code to scrape +4 months which is +1 for the next month like domestic transactions +3 for the special abroad terms...

I've tested this code both with `max` and `isracard` which are the only two usages.

### DIFF
--- a/src/helpers/dates.ts
+++ b/src/helpers/dates.ts
@@ -1,12 +1,12 @@
 import moment, { Moment } from 'moment';
 
-export default function getAllMonthMoments(startMoment: Moment | string, scrapeXFutureMonths?: number) {
+export default function getAllMonthMoments(startMoment: Moment | string, futureMonths?: number) {
   let monthMoment = moment(startMoment).startOf('month');
 
   const allMonths: Moment[] = [];
   let lastMonth = moment().startOf('month');
-  if (scrapeXFutureMonths && scrapeXFutureMonths > 0) {
-    lastMonth = lastMonth.add(scrapeXFutureMonths, 'month');
+  if (futureMonths && futureMonths > 0) {
+    lastMonth = lastMonth.add(futureMonths, 'month');
   }
   while (monthMoment.isSameOrBefore(lastMonth)) {
     allMonths.push(monthMoment);
--- a/src/helpers/dates.ts
+++ b/src/helpers/dates.ts
@@ -1,12 +1,12 @@
 import moment, { Moment } from 'moment';
 
-export default function getAllMonthMoments(startMoment: Moment | string, includeNext: boolean) {
+export default function getAllMonthMoments(startMoment: Moment | string, scrapeXFutureMonths?: number) {
   let monthMoment = moment(startMoment).startOf('month');
 
   const allMonths: Moment[] = [];
   let lastMonth = moment().startOf('month');
-  if (includeNext) {
-    lastMonth = lastMonth.add(4, 'month');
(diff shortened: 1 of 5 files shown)


esakal:
@erikash  did you consider instead extending the scraper options with optional argument like `scrapeXFutureMonths?: number` [here](https://github.com/eshaham/israeli-bank-scrapers/blob/67c29169717342326fc8df6934cb070046cf6d7a/src/scrapers/base-scraper.ts#L33). Obviously you can find a better name.

Then you can change the signature of the method `getAllMonthMoments(startMoment: Moment | string, includeNext: boolean)` to something like that
```
export default function getAllMonthMoments(startMoment: Moment | string, scrapeXFutureMonths?: number) {
 let monthMoment = moment(startMoment).startOf('month');

  const allMonths: Moment[] = [];
  const addMonthNumber = scrapeXFutureMonths || 1;
  const lastMonth = moment().startOf('month').add(addMonthNumber, 'month');

  while (monthMoment.isSameOrBefore(lastMonth)) {
    allMonths.push(monthMoment);
    monthMoment = moment(monthMoment).add(1, 'month');
  }

  return allMonths;
}
```
wdyt?

esakal:
@erikash hi did you see my comment above?

erikash:
Hi,
Sorry for the late reply...
Busy times 😉

I like your suggestion, it's the more flexible solution :)

In any case I think the `4` instead of `1` change that I suggested is a
more sane default to the existing code as well...

I'll try to find some time later this week to change the code per
your request.

Thanks,
Erik.

erikash:
@esakal I've updated the code per your suggestion.
Keep in mind that I've set the default to `2` to also fetch transactions from the month after next instead of `4` as I did originally...
I'm guessing that `2` is the most common option which would fit regular and "business cards".

esakal:
@erikash  what I don't understand is how should it works when max UI shows only one month ahead?

CleanShot 2021-11-13 at 17 32 38@2x

erikash:
@esakal my guess it that the UI shows you only one month ahead because you don't have any transactions to be processed two months from now (in January).

BTW: Me neither.. but that's only because I asked them deliberately to process my international transactions up to 3 days.
The default was +3 for my card (Max Corporate)

When I debugged it locally the API call returned `results: null` which is handled correctly by the scraper:
image

I think it's a better default anyway than just one month ahead but I'm open to changing the default back to `1` month ahead too :)

WDYS?

esakal:
@erikash because their UI doesn't let more then one month ahead, I'm not sure if we want to ask for more. It might signal on their side on suspicious activity and raise alerts of undesired behavior on their side.

Setting the default value to the same as today is a bit faster as we can merge it immediately. Asking for a month ahead by default instead of changing it to two doesn't change the current API and we just extend it with additional option thus it doesn't require additional step. 

But I also don't mind re-thinking about the default and ask @eshaham to quickly review it and comment about it.

erikash:
@esakal , It's ok :)
I've changed the default to `1`, I agree it's safer.

esakal (approved):

esakal (approved):

PR #623, opened 2021-11-05
Baruchiro/Mizrahi-Support-multiple-accounts

### LINKED ISSUE #622: Mizrahi: Support multiple accounts
If you have multiple accounts in Mizrahi, it will fetch only the first account.

image

image

baruchiro:
Fixes #622

### DIFF
--- a/src/helpers/elements-interactions.ts
+++ b/src/helpers/elements-interactions.ts
@@ -1,4 +1,5 @@
 import { Frame, Page } from 'puppeteer';
+import { waitUntil } from './waiting';
 
 async function waitUntilElementFound(page: Page | Frame, elementSelector: string,
   onlyVisible = false, timeout = 30000) {
@@ -7,6 +8,22 @@ async function waitUntilElementFound(page: Page | Frame, elementSelector: string
 
 async function waitUntilElementDisappear(page: Page, elementSelector: string, timeout = 30000) {
   await page.waitForSelector(elementSelector, { hidden: true, timeout });
+}
+
+async function waitUntilIframeFound(page: Page, framePredicate: (frame: Frame) => boolean, description = '', timeout = 30000) {
+  let frame: Frame | undefined;
+  await waitUntil(() => {
+    frame = page
+      .frames()
+      .find(framePredicate);
+    return Promise.resolve(!!frame);
+  }, description, timeout, 1000);
+
+  if (!frame) {
+    throw new Error('failed to find iframe');
+  }
+
+  return frame;
 }
 
 async function fillInput(pageOrFrame: Page | Frame, inputSelector: string, inputValue: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/helpers/elements-interactions.ts
```diff
@@ -9,6 +10,24 @@ async function waitUntilElementDisappear(page: Page, elementSelector: string, ti
...
+      .find(framePredicate);
+    return Promise.resolve(!!frame);
+  }, 'wait for iframe with login form', timeout, 1000);
+
+  if (!frame) {
+    throw new Error('failed to find iframe');
+  }
+
+  return frame;
+}
```
baruchiro, line 29: - I don't know if it should be implemented here or in `waiting.ts`.
- I think it is convenient to combine the wait and the getElement.
- The implementation is copied from `visa-cal` and can be used there too.

src/scrapers/mizrahi.ts
```diff
@@ -154,7 +157,30 @@ class MizrahiScraper extends BaseScraperWithBrowser {
...
+        success: true,
+        accounts: results,
+      };
+    } catch (e) {
+      return {
+        success: false,
+        errorType: ScraperErrorTypes.Generic,
+        errorMessage: (e as Error).message,
+      };
+    }
```
baruchiro, line 178: Note it means "everything or nothing".
Do you want me to `try-catch` each account instead?

For example, I needed to fix the pending transactions that didn't work, because it blocked the next account.

src/helpers/elements-interactions.ts
```diff
@@ -9,6 +10,22 @@ async function waitUntilElementDisappear(page: Page, elementSelector: string, ti
...
 }
 
+async function waitUntilIframeFound(page: Page, framePredicate: (frame: Frame) => boolean, timeout = 30000) {
+  let frame: Frame | undefined;
+  await waitUntil(() => {
+    frame = page
+      .frames()
+      .find(framePredicate);
+    return Promise.resolve(!!frame);
+  }, 'wait for iframe with login form', timeout, 1000);
```
esakal, line 20: consider getting the description from arg or use generic message instead like `wait for iframe`

src/helpers/elements-interactions.ts
esakal: we can keep it here for now imo.

src/scrapers/mizrahi.ts
esakal: 👍  I believe it should be all or nothing because otherwise the user might get partial information without any notice about it

esakal (changes requested):
@baruchiro I added minor comments beside them the code seems fine.

I cannot run it thought as I don't have an account in Mizrahi but I believe you ran it multiple times?

src/helpers/elements-interactions.ts
baruchiro: Done

esakal (approved):
@baruchiro I'll merge it soon. Again I couldn't test Mizrahi scrapers but I believe you tested numerous times :)

baruchiro:
> @baruchiro I'll merge it soon. Again I couldn't test Mizrahi scrapers but I believe you tested numerous times :)

Sorry, I forgot to answer.

Yes, I tested it, I needed to debug it multiple time because of the problem with the pending transactions.

Issue #627, opened 2021-11-12
[visaCal] Foreign currency parsed as NaN


sushi30:
### Output

```json
{
    "type": "normal",
    "status": "completed",
    "date": "2021-09-24T00:00:00.000Z",
    "processedDate": "2021-12-02T00:00:00.000Z",
    "originalAmount": null,
    "originalCurrency": "46.28",
    "chargedAmount": -175.54,
    "chargedCurrency": "ILS",
    "description": "AIRBNB * HMMKENEB22",
    "memo": "עסקה עם הנחה בעמלת עסקאות מט\"ח"
  }
```

### Expected

```json
{
  ...,
  "originalAmount": -46.28,
  "originalCurrency": "EUR",  # or "€"
}
```

I suspect this is because in `getAmountData`, parsing should be done in reverse. I will file a PR for this later this weekend.

https://github.com/eshaham/israeli-bank-scrapers/blob/6301efd1c1178378774a0c65fff421522ec7714d/src/scrapers/visa-cal.ts#L103-L105

### Original Transaction from CAL website
![image](https://user-images.githubusercontent.com/26347589/141466792-e5be1497-7c5c-4cfe-b345-da245685b993.png)

PR #628, opened 2021-11-12
[visaCal] Parse foreign currencies correctly

sushi30:
1. Added explicit parsing for EUR
2. Fixed parsing of amount string scraped from visa cal website.

closes https://github.com/eshaham/israeli-bank-scrapers/issues/627

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -104,7 +104,7 @@ function getAmountData(amountStr: string) {
     currency = EURO_CURRENCY;
   } else {
     const parts = amountStrCln.split(' ');
-    currency = parts[0];
+    [currency] = parts;
     amount = -parseFloat(parts[1]);
   }
 
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -13,7 +13,7 @@ import {
 } from '../transactions';
 import { ScaperOptions, ScaperScrapingResult, ScraperCredentials } from './base-scraper';
 import {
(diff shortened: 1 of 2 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -99,10 +99,13 @@ function getAmountData(amountStr: string) {
...
     amount = -parseFloat(amountStrCln.replace(DOLLAR_CURRENCY_SYMBOL, ''));
     currency = DOLLAR_CURRENCY;
+  } else if (amountStrCln.includes(EURO_CURRENCY_SYMBOL)) {
+    amount = -parseFloat(amountStrCln.replace(EURO_CURRENCY_SYMBOL, ''));
+    currency = EURO_CURRENCY;
   } else {
     const parts = amountStrCln.split(' ');
-    amount = -parseFloat(parts[0]);
-    [, currency] = parts;
+    [currency] = parts;
```
esakal, line 107: I see that you added the euro currency. But why did you changed the fallback behavior and switched places? It was there for a long time now so either it never worked or maybe you are using a different UI layout (maybe left-to-right)?

esakal (changes requested):

src/scrapers/visa-cal.ts
sushi30: Switching amount and currency worked for me even without the EURO fallback (as opposed to the original code). But maybe its not robust enough. I noticed a [library ](https://www.npmjs.com/package/parse-money) for handling the parsing of currency strings. I think in any case the method should be more robust then assuming the string will be in a certain format. The CC companies are definitely not trying to keep there "API" consistent.

esakal (approved):

PR #635, opened 2021-12-01
support empty pages

### LINKED ISSUE #633: Visa cal issues with an account that has multiple cards
Hi,
I've just upgraded to the latest version, and everything works except that account.
It's a visa-cal account with 4 (!) different cards. some of them are not active, but some are.
I was getting the following error:
`Error: failed to find element matching selector "[id$=FormAreaNoBorder_FormArea_ctlMainToolBar_lblCaption]"`
I also have another visa-cal account with just a single card, scraping works well there.
Please let me know if you need further info

esakal:
# Motivations
support pages without transactions
CleanShot 2021-12-01 at 23 17 33

closes #633

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -173,6 +173,7 @@ async function fetchTransactionsForAccount(page: Page, startDate: Moment, accoun
   const buttonSelector = '[id$="FormAreaNoBorder_FormArea_ctlSubmitRequest"]';
   const nextPageSelector = '[id$="FormAreaNoBorder_FormArea_ctlGridPager_btnNext"]';
   const billingLabelSelector = '[id$=FormAreaNoBorder_FormArea_ctlMainToolBar_lblCaption]';
+  const noDataSelector = '[id$=FormAreaNoBorder_FormArea_msgboxErrorMessages]';
 
   debug('find the start date index in the dropbox');
   const options = await pageEvalAll(page, '[id$="FormAreaNoBorder_FormArea_clndrDebitDateScope_OptionList"] li', [], (items) => {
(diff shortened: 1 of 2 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -194,60 +195,71 @@ async function fetchTransactionsForAccount(page: Page, startDate: Moment, accoun
...
       clickButton(page, buttonSelector),
     ]);
-    debug('find the billing date');
-    const billingDateLabel = await pageEval(page, billingLabelSelector, '', ((element) => {
-      return (element as HTMLSpanElement).innerText;
+    debug('check if month has no transactions');
+    const pageHasNoTransactions = await pageEval(page, noDataSelector, false, ((element) => {
+      const siteValue = ((element as HTMLSpanElement).innerText || '').replace(/[^ א-ת]/g, '');
+      return siteValue === 'לא נמצאו נתונים';
     }));
```
daniel-hauser, line 202: @esakal, Note that `$eval` (used inside `pageEval`) [throws](https://pptr.dev/#?product=Puppeteer&version=v12.0.1&show=api-frameevalselector-pagefunction-args-1) if no element matches the selector causing the scraper to fail when a card **has** transactions.

Adding a `catch` that resolves to false will fix it.

```suggestion
    })).catch(() => false);
```

esakal: @daniel-hauser did it fail to you? because the infrastructure already take care on that here https://github.com/eshaham/israeli-bank-scrapers/blob/243da809308191203796f307d910a03678c7a9dc/src/helpers/elements-interactions.ts#L87

Let me know?
Eran

daniel-hauser: Yes, it failed with `Error: failed to find element matching selector "[id$=FormAreaNoBorder_FormArea_msgboxErrorMessages]"`.

You can ignore my comment as I now see you fixed the infra error test (from `elements` to `element`) in this PR.

eshaham (approved):

PR #636, opened 2021-12-02
Add category support

daniel-hauser:
## Added
- Add support for an optional category for credit card transactions.
- Add categories for max transactions

## Accounts supported
- **max**: Added support

## Accounts not supported
The following accounts are accounts I have access to, but categories are not loaded with the transactions, only one by one when expanding in the UI.
- **visaCal**
- **amex**
- **isracard**

If you think its fine to call the get category endpoint for every transaction - I can try to add support for them.


esakal:
@daniel-hauser Hi, thanks for the contribution and detailed description. I believe it is fine with supporting only the max for now. As this is additional information added to the transaction I don't see any reason not to add it.

@eshaham I'm going to approve the PR. If you have any concerns please comment.

esakal (approved):
I will let @eshaham  to review as well before I merge it

PR #638, opened 2021-12-05
support long date format in otsar hahayal

### LINKED ISSUE #637: Otsar Hahayal bank transactions date is being written as last year instead of current year
I'm not sure if its an issue in israeli bank  scrapers but I'm using caspion to scrape my financial data and the dates for bank otsar hahayal are last year instead of this year. can anyone verify it?

esakal:
closes #637


esakal:
@eshaham how do you want to proceed? @ReaBar cannot test it. I believe the change is safe but up to you. Can you review & approve?

@brafdlog maybe you can test it?

eshaham (approved):
looks legit to me

brafdlog:
@esakal Thanks for the fix. I don't have otsar hahayal. We will update the version in caspion so @reaBar will be able to test

PR #643, opened 2022-01-01
fix: handle change password of leumi

esakal:
Handle change password scenario

CleanShot 2022-01-01 at 22 21 25@2x


esakal:
@eshaham can you please review?

esakal:
@eshaham please review and merge

Issue #655, opened 2022-02-12
Is there a way to get the future bill for credit cards such as Cal?


LiranBri:
Is there a way to get the future bill (debt) for credit cards such as Cal? or only transactions are currently supported?

LiranBri:
I see in code now that there isn't. I can contribute a PR but how would you like me to structure it in the response object? as there are no matching fields currently for it.

probably we need new fields such as:
```js
futureDebits: [
   { bankAccount: "612-12345", amount: 143, when: "2020-03-01" },
   { bankAccount: "612-12345", amount: 143, when: "2020-03-01" },
]
```

esakal:
I have merged the PR but the magic link was incorrect so I'll close this one manually

PR #656, opened 2022-02-12
feat: CAL future debits

LiranBri:
contribution following issue https://github.com/eshaham/israeli-bank-scrapers/issues/655

### DIFF
--- a/src/scrapers/visa-cal.test.ts
+++ b/src/scrapers/visa-cal.test.ts
@@ -50,19 +50,4 @@ describe('VisaCal legacy scraper', () => {
     // expect(result?.accounts?.length).toEqual(2)
     exportTransactions(COMPANY_ID, result.accounts || []);
   });
-
-  maybeTestCompanyAPI(COMPANY_ID)('should scrape future debits"', async () => {
-    const options = {
-      ...testsConfig.options,
-      companyId: COMPANY_ID,
-    };
-
-    const scraper = new VisaCalScraper(options);
-    const result = await scraper.scrape(testsConfig.credentials.visaCal);
-    expect(result).toBeDefined();
-    const error = `${result.errorType || ''} ${result.errorMessage || ''}`.trim();
-    expect(error).toBe('');
-    expect(result.success).toBeTruthy();
-    expect(result.futureDebits?.length).toBeGreaterThan(0);
-  });
 });
--- a/src/scrapers/visa-cal.test.ts
+++ b/src/scrapers/visa-cal.test.ts
@@ -50,4 +50,19 @@ describe('VisaCal legacy scraper', () => {
     // expect(result?.accounts?.length).toEqual(2)
     exportTransactions(COMPANY_ID, result.accounts || []);
   });
+
+  maybeTestCompanyAPI(COMPANY_ID)('should scrape future debits"', async () => {
+    const options = {
+      ...testsConfig.options,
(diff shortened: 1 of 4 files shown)


src/scrapers/visa-cal.test.ts
```diff
@@ -50,4 +50,19 @@ describe('VisaCal legacy scraper', () => {
...
+      companyId: COMPANY_ID,
+    };
+
+    const scraper = new VisaCalScraper(options);
+    const result = await scraper.scrape(testsConfig.credentials.visaCal);
+    expect(result).toBeDefined();
+    const error = `${result.errorType || ''} ${result.errorMessage || ''}`.trim();
+    expect(error).toBe('');
+    expect(result.success).toBeTruthy();
+    expect(result.futureDebits?.length).toBeGreaterThan(0);
```
esakal, line 66: we can use the same test we use for transactions to check this one as well in line 52, no need to run the scrapping again. 

Also, because it is optional, I don't believe we should assume in the expect that we got results. wdyt?

esakal (changes requested):

src/scrapers/visa-cal.test.ts
LiranBri: I don't think its optional for CAL, as it is expected to always have this value.
the reason it is marked as optional, is because not all scrappers implemented it (only CAL as this is the first time its implemented in this PR)

esakal:
@LiranBri Hi, I commented in the code review, please see my comment there. Basically the code it self looks functional but it result with empty array for my acccounts. So either this data is not exposed to all users and then we should assume it is optional on the CAL level and adjust the unit-tests accordingly, or we should figure out which page exposes that information and navigate to that page prior to scraping the debits

src/scrapers/visa-cal.test.ts
LiranBri: Screen Shot 2022-02-20 at 11 06 12
that how mine is looking

LiranBri: Screen Shot 2022-02-20 at 11 06 12
that how my screen looks

regarding moving the test- sure I can move it there

esakal: Hi, @LiranBri. I double checked the settings page and I don't find a way to enable this widget that you have. So we must assume this is not mandatory and other people might not have this section like in my scenario. 

We can proceed with your changes once you remove the tests as they don't apply to anyone . Since tests are not part of the ci/cd, you can just have them on your machine locally without adding them to the source control. wdyt?

src/transactions.ts
```diff
@@ -51,3 +51,10 @@ export interface Transaction {
   installments?: TransactionInstallments;
   category?: string;
 }
+
+export interface FutureDebit {
```
esakal, line 55: since this is not part of the transactions, it shouldn't be here. you can move them to `src/scrapers/base-scraper.ts` right above `ScaperScrapingResult` interface

esakal (changes requested):

src/scrapers/visa-cal.test.ts
LiranBri: sure, i get your point. I removed it.

src/transactions.ts
LiranBri: done

esakal (approved):

PR #657, opened 2022-02-22
Fix Leumi invalid password result

uraid:
The invalid password result was previously detected by redirection to '/InternalSite/CustomUpdate/leumi/LoginPage.ASP'.
This is not the case anymore, as an error message is presented to the user instead.
Fixed by detecting the invalid password error message.

### DIFF
--- a/src/scrapers/leumi.ts
+++ b/src/scrapers/leumi.ts
@@ -30,7 +30,7 @@ function getPossibleLoginResults() {
         if (!options || !options.page) {
           throw new Error('missing page options argument');
         }
-        const errorMessage = await pageEvalAll(options.page, '.errHeader', [], (label) => {
+        const errorMessage = await pageEvalAll(options.page, '.errHeader', '', (label) => {
           return (label[0] as HTMLElement)?.innerText;
         });
 
@@ -42,7 +42,7 @@ function getPossibleLoginResults() {
         if (!options || !options.page) {
           throw new Error('missing page options argument');
         }
-        const errorMessage = await pageEvalAll(options.page, '.errHeader', [], (label) => {
+        const errorMessage = await pageEvalAll(options.page, '.errHeader', '', (label) => {
           return (label[0] as HTMLElement)?.innerText;
         });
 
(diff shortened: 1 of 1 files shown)


src/scrapers/leumi.ts
```diff
@@ -19,12 +19,24 @@ const FILTERED_TRANSACTIONS_URL = `${BASE_URL}/ChannelWCF/Broker.svc/ProcessRequ
...
-    [LoginResults.InvalidPassword]: [/InternalSite\/CustomUpdate\/leumi\/LoginPage.ASP/],
+    [LoginResults.InvalidPassword]: [async options => {
+      if (!options || !options.page) {
+        throw new Error('missing page options argument');
+      }
+
+      const errorMessage = await (0, _elementsInteractions.pageEvalAll)(options.page, '.errHeader', [], label => {
+        var _ref;
+
+        return (_ref = label[0]) === null || _ref === void 0 ? void 0 : _ref.innerText;
```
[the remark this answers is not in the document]
uraid: Yeah I see it now.
Changed and tested on my local machine, and edited the installed files.
Meant to copy the code from 'LoginResults.AccountBlocked' 1 line below it.

src/scrapers/leumi.ts
```diff
@@ -25,18 +25,18 @@ const INVALID_PASSWORD_MSG = 'אחד או יותר מפרטי ההזדהות ש
...
-        return (_ref = label[0]) === null || _ref === void 0 ? void 0 : _ref.innerText;
-      });
-      return errorMessage === null || errorMessage === void 0 ? void 0 : errorMessage.startsWith(INVALID_PASSWORD_MSG);
-    }],
+    [LoginResults.InvalidPassword]: [
+      async (options) => {
+        if (!options || !options.page) {
+          throw new Error('missing page options argument');
+        }
+        const errorMessage = await pageEvalAll(options.page, '.errHeader', [], (label) => {
```
esakal, line 33: the default value should be of the same type of the expected value, so I believe it should be `''` instead of `[]`?

uraid: Sounds about right.
Also if that's the case, the 'LoginResults.AccountBlocked' case should be changed as well.
I'll edit a fix.

esakal (approved):

mmeir11 (approved):

uraid:
It seems like CI doesn't like the defaults change to string.
How should we proceed?

esakal:
@uraid I have learned that if you re-run the CI eventually it give up :) it looks fine now.

PR #664, opened 2022-03-10
max change the popup from

roeidalm:
the code is not working on Max before
changing the closePopup id
closePopup to only class="close"

### DIFF
--- a/src/scrapers/max.ts
+++ b/src/scrapers/max.ts
@@ -258,8 +258,8 @@ class MaxScraper extends BaseScraperWithBrowser {
       fields: createLoginFields(credentials),
       submitButtonSelector: '#login-password #send-code',
       preAction: async () => {
-        if (await elementPresentOnPage(this.page, '#closePopup')) {
-          await clickButton(this.page, '#closePopup');
+        if (await elementPresentOnPage(this.page, '.close')) {
+          await clickButton(this.page, '.close');
         }
         await clickButton(this.page, '.personal-area > a.go-to-personal-area');
         await waitUntilElementFound(this.page, '#login-password-link', true);


src/scrapers/max.ts
```diff
@@ -258,8 +258,8 @@ class MaxScraper extends BaseScraperWithBrowser {
       fields: createLoginFields(credentials),
       submitButtonSelector: '#login-password #send-code',
       preAction: async () => {
-        if (await elementPresentOnPage(this.page, '#closePopup')) {
-          await clickButton(this.page, '#closePopup');
+        if (await elementPresentOnPage(this.page, '.close')) {
```
esakal, line 261: I ran it locally and for me it manages to scrape with a popup that appears. But they might have various popup types.

If you are sure this is needed, you can keep both logics, they doesn't interrupt each other. so you will have something like 

```

        if (await elementPresentOnPage(this.page, '#closePopup')) {      
           await clickButton(this.page, '#closePopup');          
        }
              
        if (await elementPresentOnPage(this.page, '.close')) {      
           await clickButton(this.page, '.close');          
        }
```

esakal, line 261: I ran it locally and for me it manages to scrape with a popup that appears. But they might have various popup types.

If you are sure this is needed, you can keep both logics, they doesn't interrupt each other. so you will have something like 

```

        if (await elementPresentOnPage(this.page, '#closePopup')) {      
           await clickButton(this.page, '#closePopup');          
        }
              
        if (await elementPresentOnPage(this.page, '.close')) {      
           await clickButton(this.page, '.close');          
        }
```

esakal (changes requested):

esakal:
I'll close this PR for now. @roeidalm if it is still relevant please reply to the comment in this PR

PR #666, opened 2022-03-19
feat(mizrahi-scraper): add update email page support

naorpeled:


src/scrapers/mizrahi.ts
```diff
@@ -66,7 +66,7 @@ function createLoginFields(credentials: ScraperCredentials) {
 
 function getPossibleLoginResults(page: Page): PossibleLoginResults {
   return {
-    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL],
+    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL, async () => !!(await page.$x('//title[contains(., "עדכון כתובת דוא\'\'ל")]'))],
```
naorpeled, line 69: Would add an XPath expression generation helper to increase readability and ease of use for cases when there are several items within a contains clause for example.
Not in the scope of this PR though.

src/scrapers/mizrahi.ts
```diff
@@ -66,7 +69,7 @@ function createLoginFields(credentials: ScraperCredentials) {
 
 function getPossibleLoginResults(page: Page): PossibleLoginResults {
   return {
-    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL],
+    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL, async () => !!(await page.$x(`//title[contains(., "${emailUpdatePageHebrewTitle}") or contains(., "${emailUpdatePageEnglishTitle}")]`))],
```
naorpeled, line 72: Would add an XPath expression generation helper to increase readability and ease of use for cases when there are several items within a contains clause for example.
Not in the scope of this PR though.

esakal: Hi, thanks for the contribution. Just to verify that I understand the reason behind your change. From the code you have scenario that you are not navigated to `AFTER_LOGIN_BASE_URL` and instead you reach an other page (or maybe the same one as the login) which should the update email title. And although you see that popup, you are already logged so you can proceed with the scrapping?

naorpeled: Yes,
it's a page within the dashboard,
you can proceed with doing any actions without updating your email.

PR #681, opened 2022-04-30
fix: 🐛 Isracard scraper not working

shaharkazaz:
* Typo ScraperOptions interface
* Expose ScraperOptions interface
* Update puppeteer to v5.5.0
* Update node and npm versions to supported versions ( node < 14 is deprecated)
* Replace waitFor usages with waitForTimeout to remove console warnings

✅  Closes: #659


esakal:
Hello, Thanks you for the awesome PR. 

few questions.
1. can you update the .nvmrc file with 14.14.0 version

2. It seems that in the upgraded puppeteer the place of the revision.js file is under puppeteer so you need to change https://github.com/eshaham/israeli-bank-scrapers/blob/3347fbeb4c1f851e6c31a88c7f410f9a0c409564/utils/update-puppeteer-config.js#L11 this line and add puppeteer.
CleanShot 2022-04-30 at 15 52 26@2x

3. it still didn't work when I ran the tests but maybe it is something local I will check it out in the evening.

src/scrapers/base-isracard-amex.ts
```diff
@@ -283,6 +283,17 @@ class IsracardAmexBaseScraper extends BaseScraperWithBrowser {
   }
 
   async login(credentials: ScraperCredentials): Promise<ScaperScrapingResult> {
+    await this.page.setRequestInterception(true);
+
+    this.page.on('request', (request) => {
+      // This request is blocking the page load intentionally.
+      if (request.url().endsWith('detector-dom.min.js')) {
```
daniel-hauser, line 290: `endsWith` relies on them not to add junk at the end, consider using `includes` instead 
```suggestion
      if (request.url().includes('detector-dom.min.js')) {
```

shaharkazaz:
@esakal 
1. Done, also updated `.node-version` and the GitHub workflow.
2. Done, when I built the lib locally I made the change but I didn't quite understand the build process so I didn't want to commit it.
3. That's correct, I'm intentionally blocking this file, from what I'm seeing it takes around 45sec for this request to finish loading but it has no effect on the page itself nor do we need it to scrape the data.
4. Consider adding a "build" stage to the ci.

esakal:
@shaharkazaz I will try again tomorrow. I believe there is a problem with Isracard servers. I'm getting invalid password, or general error like this
CleanShot 2022-04-30 at 22 29 50@2x

So I cannot confirm the fix.

I believe that I should deploy your fix in two phases. To begin with I'll deploy only the fix for isracard that ignores the request to the dom-detector. After that I'll be able to merge your PR as a breaking changes. I must do it as breaking changes because it updates both the puppeteer version and the node version. 

I can understand why you would like to upgrade Node as ours is not maintained any more but:
1. what are the reasons to upgrade to current LTS v16?
2. what are the reasons to upgrade the version of the puppeteer?

shaharkazaz:
@esakal 
1. I tested it locally several times in the project I'm working on and it seemed ok. I think there might be an issue as you say with the Isracard servers.
2. I don't think that the puppeteer is a breaking change since it's a minor update and not a major version. I made the update to resolve the console log errors, the waitForTimeout was introduced in [v5.3.0](https://github.com/puppeteer/puppeteer/releases/tag/v5.3.0).
3. The node upgrade isn't mandatory. as a dev wanting to contribute to the project, I saw that the versions that it was based on are deprecated that's all 🙂 I can revert the upgrade. In general, this project needs to a major upgrade.

esakal:
@shaharkazaz  thanks again for providing this fix that fast. As I wrote in #683 I believe we should provide a solution to the Isracard asap without introducing breaking changes and address the other changes separately. 

If you agree about this approach you can either create a new PR or just edit the name to address only dependencies upgrade.

# Reasons I believe we should handle it separately 
- Changing puppeteer version and Node.JS requires people to adjust their own code base / solutions.
- It requires testing from the community to verify that all the scrapers we have are working
- Some additional tests for the code-less (core) deployment 

# Questions
- if we anyhow upgrading Node.JS, shouldn't we go to v16?
- same question for puppeteer as version 5.1.0 is deprecated and it is two years old.

shaharkazaz:
@esakal, I wanted to make minimum changes in the dependencies. So that's why I updated puppeteer from 5.1 to 5.5 (which shouldn't affect the existing project as it **not** a major version) so I can remove the warnings.

Not all people are upgrading to the latest node version.
Even if the LTS is v16, v14 is still maintained, and therefore it's the bare minimum I think it should require.
Personally, that's how I do it. As long as I don't need something from a specific version, I support the oldest **maintained** version.

As I stated before, this repo needs a **major upgrade**, both in dependencies, build process, versioning, release process, and generally in the code (from what I saw)

Unfortunately, I don't have the time to do such a project.

 I will close this PR so you can make the changes you find fit 👍

esakal:
Hi, thanks for the suggestions, I agree about many of them. 

Although this library is quite popular, mainly because there aren't other open-sources that allow scraping so many institutes, many people don't have time to actively contribute to it.

I'm waiting for the Open Banking to kick-in so we will be able to use proper APIs and deprecate this library. Till then even that major upgrade might have positive impact, I believe what we have is good enough for now. 

But I do believe that we should not use deprecated Node.JS version.  In general I also favor the approach of changing to bare minimum and although v16 seems relevant as well I'm ok with v14.

Regarding puppeteer I get your point and obviously it is safer to upgrade to the closest version possible as we don't have proper e2e tests.

As your PR already address those two things, can you please just pull from eshaham master and reopen the PR? I'll then merge it with the updates of Node.JS and the puppeteer.

shaharkazaz:
@esakal I updated the PR

esakal (approved):

esakal:
@shaharkazaz 👍  thanks!

esakal:
@eshaham please review my previous comment and help merging this one

eshaham:
@esakal done 👍

PR #687, opened 2022-05-12
chore: Add identifier to max

daniel-hauser:


src/scrapers/max.ts
```diff
@@ -124,11 +127,11 @@ function getTransactionType(txnTypeStr: string) {
 
 function getInstallmentsInfo(comments: string) {
   if (!comments) {
-    return null;
```
esakal, line 127: Just wondering why you prefer to return undefined instead of null?

daniel-hauser: `getInstallmentsInfo` was called with `|| undefined` which is basically converts null to undefined - so it's more straight forward to just return `undefined`

daniel-hauser:
@esakal Can you take a look at this PR?

esakal:
@daniel-hauser I'm having technical difficulties merging it probably because it was opened before we updated the node versions. 

Two options:
1. @eshaham can you force merge it?
2. @daniel-hauser can you try to merge master into this PR

daniel-hauser:
This branch is ahead of master.
Seems like the problem is that the node 10/12 test actions are still required by the branch protection rules - and all future PRs are blocked by it.

The fix is easy but requires access to the branch settings

eshaham:
I thought I already did... 🤦‍♂️ 
Anyway, done now.

PR #689, opened 2022-06-08
use better selector to detect page navigation in leumi

esakal:


esakal:
@eshaham can you please merge this one as well.

eshaham (approved):

dimdimi4:
So, it appears that in result of this PR I get error like this: 
```
{"success":false,"errorType":"GENERIC","errorMessage":"waiting for selector `a[title=\"דלג לחשבון\"]` failed: 
 »   timeout 60000ms exceeded"}
```

esakal:
@dimdimi4 the changes did in this PR couldn't cause the issue. But reading the code the error message can be different in between runs. Basically we have a list of conditions we are looking for at that point in the scraper to detect when you logged in successfully to the page. If you dont and you got a page we are familiar with like 'invalid password' etc we detect it as well. 

In your scenario the scrapers doesn't meet any of the conditions and you get timeout after 60 seconds.

are you sure you provide the right credentials?

To try and figure it out, you should get a snapshot of whatever you are getting after the scraper complete the login process. If you have knowledge with node and you can run the tests directly from this repo. You can use `DEBUG=israeli-bank-scrapers:*` to run the scrapers with logs but i'm not sure if it will reveal useful information as we know where you are failing. 

But once running the tests from this project you will get a snapshot of the browser when it fails under 'snapshot' folder. see https://github.com/eshaham/israeli-bank-scrapers/blob/9ef15c0d729b4ba06ac3aedf3becf0ca6e6728b3/src/tests/.tests-config.tpl.js#L13, if you do please share here the screen you are getting. 

you can also try to set `showBrowser` to `true` and see the scraper browser in action.

Please reply with the snapshot or some logs so i'll be able to try and assist more.

PR #673, opened 2022-04-17
fix(mizrahi-scraper): update selectors to make it work

naorpeled:
credits to @Asafb26 and @dumsky

closes #669

PR #710, opened 2022-08-15
Mizrahi waiting for something not happening

baruchiro:
My tests are failing on this line, and after removing it they pass, I don't remember for which workaround we need it.

I considered doing a race, to wait for the networkidle but limit it in time and continue. WDYT? And how can I do that?

Anyway, I did this work because Mizrahi failed on my "caspion", but it failed in another step, so first let's solve this and after that I will check if the problem in "caspion" still exists.

### DIFF
--- a/src/helpers/waiting.ts
+++ b/src/helpers/waiting.ts
@@ -2,6 +2,9 @@
 export class TimeoutError extends Error {
 
 }
+
+export const SECOND = 1000;
+
 function timeoutPromise(ms: number, promise: Promise<any>, description: string) {
   const timeout = new Promise((_, reject) => {
     const id = setTimeout(() => {
@@ -33,4 +36,10 @@ export function waitUntil(asyncTest: () => Promise<any>, description = '', timeo
     wait();
   });
   return timeoutPromise(timeout, promise, description);
+}
+
+export function raceTimeout(ms: number, promise: Promise<any>) {
+  return timeoutPromise(ms, promise, 'timeout').catch((err) => {
+    if (!(err instanceof TimeoutError)) throw err;
+  });
 }
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -191,8 +191,6 @@ class MizrahiScraper extends BaseScraperWithBrowser {
   }
 
   private async fetchAccount() {
-    // workaround for situations where Mizrahi pops up pages (e.g. email update page)
-    await this.page.waitForNavigation({ waitUntil: 'networkidle2' });
     await this.page.$eval(`a[href="${OSH_PAGE}"]`, (el) => (el as HTMLElement).click());
     await waitUntilElementFound(this.page, `a[href="${TRANSACTIONS_PAGE}"]`);
     await this.page.$eval(`a[href="${TRANSACTIONS_PAGE}"]`, (el) => (el as HTMLElement).click());
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -6,6 +6,7 @@ import {
 } from '../helpers/elements-interactions';
 import { fetchPostWithinPage } from '../helpers/fetch';
 import { waitForUrl } from '../helpers/navigation';
+import { raceTimeout, SECOND } from '../helpers/waiting';
 import {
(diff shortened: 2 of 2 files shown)


esakal:
@baruchiro it was added as part of #673 by @naorpeled to support the new Mizrahi site. From the comment it is clear that it should help with some use-cases so I wouldn't just remove it with double checking it or providing a more comprehensive solution.

@naorpeled can you check this PR and see if it works for you as well without this line? or maybe share some information about when it is needed.

@baruchiro  your suggestion to support both timeout and this one might be good enough. You can do something similar to this function [timeoutPromise](https://github.com/eshaham/israeli-bank-scrapers/blob/849f6d0497cd2829f2b8cb4f289d0f1138b3accc/src/helpers/waiting.ts#L5) but instead of reject just resolve.

@naorpeled @baruchiro wdyt?

naorpeled:
@esakal will look into it :)

naorpeled:
@esakal @baruchiro without that line the scraper fails when I test it on my account.

I have an "update my email" page, without waiting for the navigation to finish, the scraper wouldn't be able to find the osh page navigation button and click it.

I created a PR that potentially can resolve the situation you're encountering,
maybe what's happening is that we're not waiting enough for the network requests to reduce.

P.S: also tried different methods to make it work, but this one seems to be the cleanest and fastest

esakal:
@naorpeled I saw the PR you provided and I'm not sure it will resolve the issue.

@naorpeled if this update my email is not part of the common flow and due to that line it fails to the majority of the users, we should remove that line temporary and figure out how to support it. 

1. @naorpeled Can you please share a snapshot of that `update my email` page?
2. that line is in `fetchAccount` method which is in a `for` loop which sounds like not the right place, but I never saw it. @baruchiro  @naorpeled how many accounts do you have available in the page, 1 or more?
3. @baruchiro @naorpeled  did you try my suggestion in a previous comment: 
> @baruchiro your suggestion to support both timeout and this one might be good enough. You can do something similar to this function [timeoutPromise](https://github.com/eshaham/israeli-bank-scrapers/blob/849f6d0497cd2829f2b8cb4f289d0f1138b3accc/src/helpers/waiting.ts#L5) but instead of reject just resolve.
 
I believe the suggestion might fit you both, even that it is an ugly workaround. But scarping is ugly sometimes the important question is if it works or not. 

wdyt?

baruchiro:
@esakal I updated my change, it is still working.

I tested the @naorpeled suggestion in #713 but it didn't work.

naorpeled:
> @naorpeled I saw the PR you provided and I'm not sure it will resolve the issue.
> 
> @naorpeled if this update my email is not part of the common flow and due to that line it fails to the majority of the users, we should remove that line temporary and figure out how to support it.
> 
> 1. @naorpeled Can you please share a snapshot of that `update my email` page?
> 2. that line is in `fetchAccount` method which is in a `for` loop which sounds like not the right place, but I never saw it. @baruchiro  @naorpeled how many accounts do you have available in the page, 1 or more?
> 3. @baruchiro @naorpeled  did you try my suggestion in a previous comment:
> 
> > @baruchiro your suggestion to support both timeout and this one might be good enough. You can do something similar to this function [timeoutPromise](https://github.com/eshaham/israeli-bank-scrapers/blob/849f6d0497cd2829f2b8cb4f289d0f1138b3accc/src/helpers/waiting.ts#L5) but instead of reject just resolve.
> 
> I believe the suggestion might fit you both, even that it is an ugly workaround. But scarping is ugly sometimes the important question is if it works or not.
> 
> wdyt?

Hey,
for now let's revert that line and when I have time for it I'll create a PR that will handle this case in a better, non-breaking, way.

Will update my PR to revert it. Sounds good?

baruchiro:
Yes, sounds good to me since I don't have that `update my email` page 😄

esakal:
@naorpeled I hope it will not break your flow, if you need assistance let me know and we can do pair programming on that or whatever will help you to make it work for you as well. 

I'm going to merge @baruchiro PR per the comments above. will do it soon

baruchiro:
@esakal note that my change should work with @naorpeled 's change, because you merged the version with the `race` between the email and the timeout.

esakal:
@baruchiro yes I saw, although it is set to 5 seconds so it might not be enough. anyhow lets let @naorpeled  to play with your solution

PR #711, opened 2022-08-22
Option to set the default timeout on puppeteer's `page.setDefaultTimeout()`

erikash:
So, I've been traveling abroad and the internet is not as good.. 
I've added this handy option to increase the default timeout - this should effect all navigation and other DOM interacting APIs.


erikash:
👋 @esakal @eshaham 
Mind taking a look?

esakal (approved):

esakal:
@erikash, Sorry for not responding earlier, I'm on a long summer vacation with limited access to a computer. 

I approved it but just before merging, I'm not sure it provide a complete solution because in our utils we have several places where we set explicit timeout and I believe it should fail scraping in slow networks. wdyt?

https://github.com/eshaham/israeli-bank-scrapers/blob/8c059f3f155ebc72c813188b8d0748f10f419107/src/helpers/elements-interactions.ts#L5

If in reality it helped you we can proceed with that one, or we can maybe consider adjusting the timeouts to prioritize the minimum between the value provided and the default timeout as I believe it was added originally to allow extending the default timeout when needed per use-case

erikash:
I've removed some hardcoded timeouts - they are redundant as the default is already 30 seconds.

My changes should affect the following methods (which helped me quite a lot now that i'm abroad)
```
page.goBack()
page.goForward()
page.goto()
page.reload()
page.setContent()
page.waitFor()
page.waitForFileChooser()
page.waitForFunction()
page.waitForNavigation()
page.waitForRequest()
page.waitForResponse()
page.waitForSelector()
page.waitForXPath()
```

esakal (approved):

erikash:
Enjoy your vacation ✨
Sorry for disturbing you…

It already helped in some scrapers…
I’ll do some more research and get back to you :)

PR #712, opened 2022-08-26
Change default moment tz to `Asia/Jerusalem`.

erikash:
We have to set `moment` timezone to `Asia/Jerusalem` in case you're running the scrapers outside Israel.
Moment's default behavior is to parse a date string using the local timezone, this is problematic as most users will visualize the Date's returned in Asia/Jerusalem timezone at some point..

For example, I'm not in Korea where it's UTC+9:00, before the fix I got the dates attached in the screenshots - after the fix it's `2022-08-19T21:00:00.000Z` per Israel's UTC+3:00 IDT timezone.

I wasn't sure about the best place to put this logic as there are two `initialize` methods in which the derived class doesn't call to the parent's initialize method...

Any good reason why `base-scraper-with-browser.ts` doesn't call `super()`?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,17464 @@
 {
   "name": "israeli-bank-scrapers",
   "version": "1.0.4",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "israeli-bank-scrapers",
+      "version": "1.0.4",
+      "license": "MIT",
+      "dependencies": {
+        "build-url": "^2.0.0",
+        "core-js": "^3.1.4",
+        "debug": "^4.3.2",
+        "husky": "^4.2.5",
+        "json2csv": "^4.5.4",
+        "lodash": "^4.17.10",
+        "moment": "^2.22.2",
+        "moment-timezone": "^0.5.35",
+        "node-fetch": "^2.2.0",
+        "puppeteer": "^6.0.0",
+        "uuid": "^3.3.2"
+      },
+      "devDependencies": {
+        "@babel/cli": "^7.4.4",
+        "@babel/core": "^7.4.5",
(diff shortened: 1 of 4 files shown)


erikash:
Screen Shot 2022-08-24 at 11 53 31
rawTransaction - as scraped
Screen Shot 2022-08-24 at 11 52 34
parsed using Korea local TZ (previous behavior)
Screen Shot 2022-08-24 at 11 33 24

esakal:
@erikash great catch! I actually returning tomorrow from a month in the USA and a few days ago I realized that all my transactions are affected from timezone which corrupted my YNAB account and now I have to manually fix everything.

So your PR would prevent future issues to others.

Unfortunately I cannot merge it as-is. One of the tests fails because you manually added `moment-timezone` to `package.json` instead of running `npm i moment-timezone` thus `package-lock.js` is not in sync.

Please remove the manual addition and run `npm i ` to have the lock file updated as well and i'll be able to merge the PR.

regarding the initialize it looks like a false implementation, feel free to fix it by replacing the line `this.emitProgress(ScaperProgressTypes.Initializing);` in `base-scraper-with-browser.ts` with `super.initialize();`

erikash:
Hope you enjoyed your travels!
Yeah it took me some time to reconcile after fixing as well :)

Actually I didn’t push the lock file on purpose as I have a newer version
of npm and wasn’t sure if upgrading to the new lockfile format would break
anything…

I’ll give it a go later today :)
And will fix the ‘super ()’ call..

Have a safe flight back :)

erikash:
@esakal check it out.. hope it's good :)

esakal:
@erikash thanks :)

As you predicted the package-lock was altered completely as your node version is probably higher then .v14.14.0

As the change is relatively small, to simplify it so you will not need to switch node version just for that, i created #715

I'll merge it soon.

Thanks again

PR #713, opened 2022-08-26
fix(scrapers/mizrahi): attempt to resovle waitForNavigation crash

naorpeled:


src/scrapers/mizrahi.ts
```diff
@@ -192,7 +192,7 @@ class MizrahiScraper extends BaseScraperWithBrowser {
 
   private async fetchAccount() {
     // workaround for situations where Mizrahi pops up pages (e.g. email update page)
-    await this.page.waitForNavigation({ waitUntil: 'networkidle2' });
+    await this.page.waitForNavigation({ waitUntil: 'networkidle2', timeout: 120_000 });
```
esakal, line 195: did you meant to write the value `120_000` with underscore? can you explain what that value should do?

if the purpose is to extend the timeout then I'm not sure it will work as we have a global timeout. that will probably fail before.

See my comment in #710

baruchiro: It didn't solve the problem, it just made the browser wait **a long time** before it failed.

naorpeled: @baruchiro I see,
then for now let's revert that line and I'll think of a better method to handle this edge case :)

PR #715, opened 2022-08-30
support time-zone based on PR 712

esakal:
closes #712

A copy of the changes from #712 with an exception for the package-lock.json file which was created by node v14.14.0


esakal:
@erikash I commit this one as breaking change as it might impact people that always scrape from outside of Israel. 

i didn't add a way to customize it from options because I assume we don't have people that always scrape from aboard but if someone will open an issue we can easily add that option and default it to Asia/Jerusalem.

erikash:
Yeah I agree..
It doesn’t make much sense to be able to change the timeline neither.. as
the Israeli banks and financial institutions all operate in Jerusalem time
zone too…

It only make sense if consumers outside Israel compensate manually for the
difference which might break their code as you mentioned…

PR #718, opened 2022-09-03
fix(scrapers/mizrahi): resolve crash when update email page is present

naorpeled:
- increased raceTimeout timeout time in `fetchAccount`

### DIFF
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -193,7 +193,7 @@ class MizrahiScraper extends BaseScraperWithBrowser {
 
   private async fetchAccount() {
     // workaround for situations where Mizrahi pops up pages (e.g. email update page)
-    await raceTimeout(5 * SECOND, this.page.waitForNavigation({ waitUntil: 'networkidle2' }));
+    await raceTimeout(30 * SECOND, this.page.waitForNavigation({ waitUntil: 'networkidle2' }));
     await this.page.$eval(`a[href="${OSH_PAGE}"]`, (el) => (el as HTMLElement).click());
     await waitUntilElementFound(this.page, `a[href="${TRANSACTIONS_PAGE}"]`);
     await this.page.$eval(`a[href="${TRANSACTIONS_PAGE}"]`, (el) => (el as HTMLElement).click());


naorpeled:
@esakal @baruchiro

baruchiro:
Yes, it is working for me. Thank you!

naorpeled:
hey @esakal,
can you please approve this? 🙏 

thanks in advance!

esakal:
@baruchiro it means that you will wait every time an extra 30 seconds before the process will continue, or did I miss something?. don't mind approving it if you feel its fine. Otherwise we can maybe think on a shorter period?

@naorpeled @baruchiro wdyt?

baruchiro:
To be honest, I'm not sure, because of the `page.waitForNavigation({ waitUntil: 'networkidle2' })` is sometimes confusing, it depends on the network and it is hard to research.

This PR is OK for me. Worst case we will wait 30 seconds, I don't think this project declared a limitation for the scraper to run, and sometimes it takes time.

baruchiro (approved):

naorpeled:
> @baruchiro it means that you will wait every time an extra 30 seconds before the process will continue, or did I miss something?. don't mind approving it if you feel its fine. Otherwise we can maybe think on a shorter period?
> 
> @naorpeled @baruchiro wdyt?

in cases the page transition wouldn't finish we would need to wait 30 seconds,
I've tried to decrease the amount of seconds but for now it seems to be the sweet spot, it's worth improving this in the future but imo we should at least have something that is working in all cases even if it takes a bit longer.

esakal:
Great as long as you are both satisfy from the suggested solution. I'll merge it now

baruchiro:
Hi @naorpeled @esakal, after it worked before, now I'm getting `TimeoutError` but I can't understand why and if it is a bug or not:

image
`err instanceof TimeoutError === false`
https://github.com/eshaham/israeli-bank-scrapers/blob/962e0b40d545b6f48035b5e1826826dc6e91647a/src/helpers/waiting.ts#L41-L45

naorpeled:
> Hi @naorpeled @esakal, after it worked before, now I'm getting `TimeoutError` but I can't understand why and if it is a bug or not:
> 
> image
> `err instanceof TimeoutError === false`
> https://github.com/eshaham/israeli-bank-scrapers/blob/962e0b40d545b6f48035b5e1826826dc6e91647a/src/helpers/waiting.ts#L41-L45
> 

Not sure either tbh.

baruchiro:
@naorpeled @esakal I think it is unstable, today I managed to run this project unit tests, but still with caspion (using the latest `v2.1.1` version) it failed.

Can we remove this line until we find a better solution?

My tracking system is missing a lot of data already.

PR #720, opened 2022-09-16
fix leumi borken login flow due to site upgrade

esakal:
# Motivation
Leumi developers seems to switch to FE framework in the login page with styling framework that doesn't use id attributes. Should align the login flow to detect the right components

@erikash I'm happy for the Leumi developers that are now using a frontend framework in the login instead of (I assume) asp.net, but it is annoying that they don't let us know about changes :) Anyhow can you review and approve?


erikash:
@esakal totally agree...
Sorry but I'm no longer a Leumi customer.. Ive gone international with Beinleumi now :)

Good luck with this one!

PR #725, opened 2022-10-06
Leumi: Init milliseconds in Leumi's date

itairaz1:
For some reason Leumi's dates has 3 milliseconds (at least for my case). I just change the milliseconds to 0 in order to fix that


esakal:
Hi @itairaz1 . I'm ok with this change but as it might cause consolidation issues to people who count on the previous date values, lets wait a few days to see if anyone raises any concern before we merge it

esakal:
@brafdlog  @baruchiro hi, see my comment above, let me know if you see any concern here

PR #726, opened 2022-10-06
Isracard: use fullPaymentDate if exist for process date

itairaz1:
Currently it takes process date from the page process date, but some transactions has different process date (fullPaymentDate), in that case, pick this different process date

### DIFF
--- a/src/scrapers/base-isracard-amex.ts
+++ b/src/scrapers/base-isracard-amex.ts
@@ -46,6 +46,7 @@ interface ScrapedTransaction {
   dealSumOutbound: boolean;
   currencyId: string;
   dealSum: number;
+  fullPaymentDate?: string;
   fullPurchaseDate?: string;
   fullPurchaseDateOutbound?: string;
   fullSupplierNameHeb: string;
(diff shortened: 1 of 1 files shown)


esakal:
@itairaz1 I'm not sure if you are resolving a bug or suggesting a better field to use. As it will break people history when comparing it is important to understand if the suggested solution is necessary.

The official API exposes two fields per transaction: `date` and `processedDate`.  I believe the value we use today is correct as processed date is when it was processed by the company and billed in the bank (processed by the bank). While `date` is when the transaction happened. But in your PR you changed `processedDate` and not `date` and I'm not sure if it is correct.

@itairaz1  @baruchiro  @erikash wdyt?

itairaz1:
I'm solving a bug here.
I didn't touch `date` field, only processing date - 
The current state is that each transaction in the month view is getting the processing date of the same month. But this is not true for all cases, some transactions in this view have different processing date, like immediate debit or Shotef+. In this case `fullPaymentDate` will have value instead of null.
I validated this in [my project](https://github.com/itairaz1/israeli-bank-firefly-importer), on my own account, and the processing date in my bank account is equivalent to this logic.

erikash:
Hi @itairaz1 :)
Long time... 
I totally understand what you wanted to do as some transactions (such as cash withdrawals?) are processed by the bank a few days after creation....
Unfortunately I don't have such transactions to test this PR with... 🤷‍♂️

baruchiro:
@esakal thank you for asking us when you see breaking changes, but I think you should always consider there are more consumers, and if it is a breaking change, declare it as such, no matter if we are handling it in Caspion or not.

Anyway, in Caspion we are taking the `date` to identify the transaction, I guess it is because the `processedDate` is updating. So for us, it is OK.
https://github.com/brafdlog/caspion/blob/08a62a57a82174048b7ad1ec899659ddeccdd0fb/src/backend/transactions/transactions.ts#L17-L21

CC: @brafdlog

esakal:
Great thanks all.
I prefer asking people that I know who are actively using this project before merging things that might break the usage as we don't have a testing environments.

@baruchiro regarding the breaking change with other consumers don't worry I always increate major if it is a breaking change.

@itairaz1  thanks for clarifying it and again for the contribution!

PR #733, opened 2022-10-29
Add support to config timeout

itairaz1:
In order to support slow end devices, I added the ability to configure timeout.
No breaking change - when leaving empty (undefined) it will use the previous undefined value, which result in 30s default.

### DIFF
--- a/src/scrapers/base-scraper-with-browser.ts
+++ b/src/scrapers/base-scraper-with-browser.ts
@@ -138,6 +138,7 @@ class BaseScraperWithBrowser extends BaseScraper {
     } else {
       const executablePath = this.options.executablePath || undefined;
       const args = this.options.args || [];
+      const { timeout } = this.options;
 
       const headless = !this.options.showBrowser;
       debug(`launch a browser with headless mode = ${headless}`);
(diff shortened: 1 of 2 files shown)


esakal:
please update https://github.com/eshaham/israeli-bank-scrapers/blob/master/src/scrapers/base-scraper.ts with the additional option.

itairaz1:
@esakal I already [did](https://github.com/eshaham/israeli-bank-scrapers/pull/733/files#diff-7730c819d8713cdcaa67d6ac00233b8c315be50c4bdddc4ec9ff4877f93867fdR98). Did I miss something?

brafdlog:
@itairaz1  thats a great idea :)
This only affects timeout for the browser launching right? Not the timeout for when waiting for an element to appear etc

brafdlog:
If you want this to apply to navigation timeout you also need to  `setDefaultNavigationTimeout`

itairaz1:
@brafdlog Thanks!
It's actually puppeteer timeout, by their doc it's `Maximum navigation time in milliseconds` which means it's the timeout that waiting for the elements.
We are using the default of puppeteer, which is 30000 (30 secs)

PR #738, opened 2022-11-09
Fix hapoalim scraper fetching only first 150 transactions.

skoant:
Increase requested transaction from 150 to 1000

### DIFF
--- a/src/scrapers/hapoalim.ts
+++ b/src/scrapers/hapoalim.ts
@@ -15,6 +15,7 @@ import { getDebug } from '../helpers/debug';
 const debug = getDebug('hapoalim');
 
 const DATE_FORMAT = 'YYYYMMDD';
+const MAX_JAVA_INT = 2147483647;
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 declare namespace window {
@@ -112,8 +113,18 @@ function convertTransactions(txns: ScrapedTransaction[]): Transaction[] {
   });
 }
 
+function getStartDate(options: ScraperOptions) {
+  const maximumYearsBack = 2;
+  const defaultStartMoment = moment().subtract(maximumYearsBack, 'years').add(1, 'day');
+  const startDate = options.startDate || defaultStartMoment.toDate();
+  const startMoment = moment.max(defaultStartMoment, moment(startDate));
+
+  const startDateStr = startMoment.format(DATE_FORMAT);
+  return startDateStr;
+}
+
 async function getRestContext(page: Page) {
-  await waitUntil(() => {
+  await waitUntil(async () => {
     return page.evaluate(() => !!window.bnhpApp);
   }, 'waiting for app data load');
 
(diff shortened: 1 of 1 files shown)


src/scrapers/hapoalim.ts
```diff
@@ -15,6 +15,7 @@ import { getDebug } from '../helpers/debug';
 const debug = getDebug('hapoalim');
 
 const DATE_FORMAT = 'YYYYMMDD';
+const MAX_JAVA_INT = 2147483647;
```
esakal, line 18: this value seems very high, it might raise concern on the other side if they log the requests, why not having something more realistic like 500?

src/scrapers/hapoalim.ts
```diff
@@ -112,6 +113,16 @@ function convertTransactions(txns: ScrapedTransaction[]): Transaction[] {
   });
 }
 
+function getStartDate(options: ScraperOptions) {
+  const maximumYearsBack = 2;
+  const defaultStartMoment = moment().subtract(maximumYearsBack, 'years').add(1, 'day');
```
esakal, line 118: all the scrapers default time is 1 years, changing it here will cause a breaking change. please remove any changes not related to the original issue from this PR. Thanks :)

btw in #728 we discussed about introducing a new option to allow customization of that default start time. but that PR is also not for that.

If you want please create an **issue** about ability to customize that value and we can discuss there the API adjustments

esakal (changes requested):

src/scrapers/hapoalim.ts
skoant: Limiting to 500 will potentially lead to the same issue we try to solve here... but I see your point.
For more context: hapoalim limits number of transactions asked from the server to 150 which is actually enforced only on the client side, the api is able to return more than 150 so setting it above that might be risky as we basically exploiting a bug / feature here :)

The proper way of solving this will be to reverse enginer their "Load more" functionality which will take more effort.

src/scrapers/hapoalim.ts
skoant: This was part of the issue since 1 year was hardcoded in the startDate logic in the hapoalim scraper that's why I don't understand how that affects other scrapers and how it's a breaking change?

src/scrapers/hapoalim.ts
skoant: @esakal wdyt?

esakal: @skoant you are right, the proper way will take much more time.

I'm ok with setting higher value, but maybe lets set a more modest value like 1000 or 1500. I understand it still raise flags on their side but it looks a bit better

esakal (changes requested):

src/scrapers/hapoalim.ts
skoant: This will enable me to extend scraping to 2 years for Hapoalim but being it a global config, it will be confusing as one may think this option affect all scrapers in addition to the fact that it won't allow setting different values for each scraper.

I think that a better and a more straight forward approach will be to implement a new per scraper config level with the current startDate and defaultStartMoment values for backward compatibly.
This will enable passing different values, flags, option which are scraper specific and also refactor away the code duplicity we have now where each scraper defines it's own defaultStartMoment.
@esakal wdyt?

esakal (approved):

PR #751, opened 2023-01-14
feat(visa-cal): Support additional info to get categories #740

### LINKED ISSUE #740: [Discussion] Implement extra scrapping for category
Hi,
I've started implementing scrapping categories (Maybe more information) per transaction (currently for isracard)
This requires perform extra click for every transaction
It will work only when specific configuration will turned on.
Any insides? anyone tried doing it already?

@eshaham @esakal @erikash

itairaz1:
Passing additionalTransactionInformation will fetch ccategory also from visa

closes #740

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -139,19 +139,19 @@ function getTransactionInstallments(memo: string): TransactionInstallments | nul
 
 function getIdentifierAndNumerator(onclickValue: string | null): { identifier?: string, numerator?: string } {
   if (!onclickValue) {
-    debug('onclick attribute not found for transaction');
+    debug('cannot extract the identifier of a transaction, onclick attribute not found for transaction');
     return {};
   }
   const expectedStartValue = 'OnMouseClickRow(this, event, "';
   if (!onclickValue.startsWith(expectedStartValue)) {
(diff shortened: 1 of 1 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -126,6 +136,30 @@ function getTransactionInstallments(memo: string): TransactionInstallments | nul
     total: parseInt(parsedMemo[2], 10),
   };
 }
+
+function getIdentifierAndNumerator(onclickValue: string | null): { identifier?: string, numerator?: string } {
+  if (!onclickValue) {
+    debug('onclick attribute not found for transaction');
```
esakal, line 142: Consider adding explanation what does it mean, something like `cannot extract the identifier of a transaction, ...`

Also, does the debug messages indicate on a bug?

If it can sometimes happen then I recommend not to write them as it will clutter the logs. Lets keep them only if when happens it is a bug that must be fixed.

itairaz1: Thanks
Will add more context to the message, as suggested.
I didn't saw this cases when there is no onclick attribute (or other debug cases below), so maybe debug log will help me capture those. wdyt?

esakal (approved):

Issue #734, opened 2022-10-30
Harel and 2FA


mnavon:
Hi
I would like to write a new scraper for Harel.
But Harel login works with 2FA and sends code via SMS.
Is there a way to overcome this issue?

esakal:
Hi @mnavon this is a great question. 
I know there are some approaches to this issue, like creating a pub/sub based async communication to ask the application/service who is using our library to provide such an information.

It will change the methodology we had so far by not blocking the scrapping process, but it might be inevitable. 

@eshaham  @erikash  @itairaz1  wdyt?

erikash:
This could be an opportunity to discuss the overall design - maybe we should create a new AsyncBaseScraper and let the consuming code call the login and scrape on it's own (also to fill in the OTP)

@mnavon @esakal @itairaz1  wanna jump on a quick zoom call / huddle to discuss?

itairaz1:
Hi,
IMO, since this is lib that uses by other apps, it need to be opened to any architecture that will be implemented by the apps the uses it.
So in this case, we can do simple solution like providing a callback function, this will provide other apps to implement any approach they choose.
The callback function can be provided in the options/configs section, and if 2FA is required and no callback is provided, and error will be raise.
@mnavon @esakal @itairaz1 @erikash  WDYT?

esakal:
I believe a callback is a good options.
1. a scraper can support 2FA.
2. if reached a 2FA flow and options has a callback, **something like 'on2FAScrapping'**, it will be called in async way.
3. if the callback is not provided, **a new type of error (something like `ScraperErrorTypes.2FASupportRequested** will be returned to the implementor indicated that 2fa callback is required to complete the process.
4. in the documentation we will advice to extend the timeout period if they expect to have 2fa flow in their use-case. The timeout option is general to all scrapers but I believe it is ok
5. Edge case concern, the puppeteer will reach timeout before getting a response of the 2FA. I believe it is ok and will not create a memory leak, wdyt?

a suggestion to the `on2FAScrapping` signature:
```
on2FAScrapping?: (companyId: CompanyTypes) => Promise<{string}>
```

What do you think?

itairaz1:
@esakal Yeah... it make sense

esakal:
@mnavon feel free to introduce the 2FA concept together with the Harel scraper. You can follow the requirements in https://github.com/eshaham/israeli-bank-scrapers/issues/734#issuecomment-1299677087

Issue #706, opened 2022-07-07
Other financial institutes scrapers


doron-cohen:
Hi,
I just found out about this project, so first of all thank you for maintaining it.
I would like to ask if you know of any scrapers for other financial institutes that offer services like Pensia, Keren Hishtalmut, stock exchange investment etc. If not, do you consider supporting these?

doron-cohen:
About collaborating, I'll be glad to do so! I  can start investing time in this only after the summer vacation but let's talk.

gilgardosh:
> It's hell

you're being polite here.
amazing that you are into it and willing to get your hands dirty 😄 ping me whenever you want to dive in

erikash:
Hi guys,
I'm one of the founders of `www.mukdam.co.il`.
We're a digital insurance agency with aspirations for becoming a digital family office in the future.

I started the company for the same frustrations you're experiencing with data and tracking - I also use a google sheet - a pretty good one that sends monthly reports via email and tracks your net worth over time...

What can we offer right now?
If you nominate us as your insurance agent (will not effect any existing agreements you have with any existing pension funds, provident funds, advanced study funds), we'll be able to fetch your Masleka every month for pretty much free and provide an API which can be consumed with Integromat (Now make), Zapier or any other automation tools which can push it Google Sheets or any other app...

Sounds interesting? feel free to PM me or click the contact us via WhatsApp in our website at `www.mukdam.co.il`.

Urigo:
@erikash is any of the work that you are doing open source?

erikash:
Hi @Urigo ,
Not yet but would be happy to open source parts of it!

Urigo:
what parts?  we would love to help with that!

erikash:
Let's talk :)
I've send you an email

doron-cohen:
@erikash count me in. Maybe we can talk over the project's Discord? Or somewhere else we can collaborate?

erikash:
Sounds great @doron-cohen !
Didn't find the project's discord server.. 😬
My email is: erik@mukdam.co.il
Shoot me an email and we'll start the conversation 🙏

erikash:
@doron-cohen still waiting for that email 😅

doron-cohen:
@erikash sending one today

esakal:
@doron-cohen feel free to reopen with comment if you would like to continue this discussion

PR #760, opened 2023-02-25
feat: Introducing 2FA scrapers infrastructure + OneZero experimental scraper

orzarchi:
Hi there!
Hopefully this is the first PR in the era of 2FA scrapers.
I've loosely followed the guidelines set up [here](https://github.com/eshaham/israeli-bank-scrapers/issues/734).

1. I've created a new base scraper - `BaseTwoFactorAuthScraper`.
2. Since the original `base-scraper` file was getting a bit long, I've cleaned up a bit and moved error related types to `errors.ts`, and public facing interfaces (e.g. `ScraperOptions`, `ScraperScrapingResult`) to `interface.ts`
3. I've created the first 2FA scraper - OneZero. It's not perfect, but I've managed to scrape my personal account successfully. See the new documentation in the `README` to learn more about it.

Some limitations about OneZero:
* I've supplied a way to provide a persistent OTP token, but it's limited - You can get one from OneZero that lasts 45 minutes, and it should be possible to refresh it indefinitely, but I've still not found the refresh graphql calls.
* Using OneZero scraper in an automated scripts will probably log you out from whatever mobile devices you are connected to - their servers don't appreciate simultaneous logins. Perhaps this can be fixed in the future by supplying more headers and persisting cookies.

closes #765


esakal:
@orzarchi thank you for this contribution 💯 . I'll free sometime soon to review and play with the PR

baruchiro:
I need to understand the interface changes more.

Meanwhile, I think you can continue with this change, it will not break our app because we will have to enable the new bank manually after getting your version.

CC: @brafdlog

orzarchi:
Any news?
@baruchiro for the record, there isn't any interface changes, except exporting more types.
And even that can be reverted if necessary

brafdlog:
Hi @orzarchi 
I (and I assume baruch as well) am in a very busy period.
I suggest you don't wait for us. There shouldn't be issues based on your description, and we will update if there is when we update to the next version of israeli-bank-scrapers.

src/helpers/fetch.ts
```diff
@@ -30,21 +30,27 @@ export async function fetchGet<TResult>(url: string,
...
+  variables: Record<string, unknown> = {},
+  extraHeaders: Record<string, any> = {}): Promise<TResult> {
+  const result = await fetchPost(url, { operationName: null, query, variables }, extraHeaders);
+  if (result.errors?.length) {
+    throw new Error(result.errors[0].message);
+  }
+  return result.data as Promise<TResult>;
+}
+
+export async function fetchGetWithinPage<TResult>(page: Page, url: string): Promise<TResult | null> {
```
erikash, line 53: No await -> no need for this to be marked `async`

src/helpers/fetch.ts
```diff
@@ -30,21 +30,27 @@ export async function fetchGet<TResult>(url: string,
...
-  extraHeaders: Record<string, any>) {
-  let headers = getJsonHeaders();
-  if (extraHeaders) {
-    headers = Object.assign(headers, extraHeaders);
-  }
+  extraHeaders: Record<string, any> = {}) {
   const request = {
     method: 'POST',
-    headers,
+    headers: { ...getJsonHeaders(), ...extraHeaders },
```
erikash, line 36: Nice!

src/helpers/fetch.ts
erikash, line 53: No await -> no need for this to be marked async

src/scrapers/base-two-factor-auth-scraper.ts
```diff
@@ -0,0 +1,63 @@
...
+import { createGenericError, createTimeoutError, ErrorResult } from './errors';
+import type {
+  TwoFactorAuthScraper,
+  ScraperOptions,
+  ScraperTwoFactorAuthTriggerResult,
+  ScraperScrapingResult,
+  ScraperGetLongTermTwoFactorTokenResult, ScraperCredentials,
+} from './interface';
+
+export type ScraperLoginResult = ErrorResult | {
```
erikash, line 11: Why not move to `interface`?

src/scrapers/base-two-factor-auth-scraper.ts
```diff
@@ -0,0 +1,63 @@
...
+  success: true;
+  persistentOtpToken: string;
+};
+
+
+export class BaseTwoFactorAuthScraper<TTwoFactorCredentials> implements TwoFactorAuthScraper {
+  constructor(protected options: ScraperOptions) {
```
erikash, line 18: Love it!

src/scrapers/one-zero.ts
```diff
@@ -0,0 +1,323 @@
...
+    this.accessToken = accessToken;
+
+    return {
+      success: true,
+      persistentOtpToken: otpTokenResult.longTermTwoFactorAuthToken,
+    };
+  }
+
+  private async fetchPortfolioMovements(portfolio: Portfolio, startDate: Date): Promise<TransactionsAccount> {
+    const account = portfolio.accounts[0];
```
erikash, line 220: Please open a task for future multi account support

src/scrapers/one-zero.ts
```diff
@@ -0,0 +1,323 @@
...
+    let cursor = null;
+    const movements = [];
+
+
+    while (!movements.length || new Date(movements[0].movementTimestamp) >= startDate) {
```
erikash, line 225: Use `moment` and not `Date` to avoid timezone issues

src/scrapers/one-zero.ts
```diff
@@ -0,0 +1,323 @@
...
+          );
+
+      movements.unshift(...newMovements);
+      cursor = pagination.cursor;
+      if (!pagination.hasMore) {
+        break;
+      }
+    }
+
+    movements.sort((x, y) => new Date(x.movementTimestamp).valueOf() - new Date(y.movementTimestamp).valueOf());
```
erikash, line 250: Same here

src/scrapers/one-zero.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    const matchingMovements = movements.filter((movement) => new Date(movement.movementTimestamp) >= startDate);
```
erikash, line 252: And here

erikash (commented):
Great work :)
Please see my comments

esakal:
@orzarchi @erikash  Hi, 
Thanks again for this one, supporting otp is required as many services switch to OTP login.

# TL;DR
Long time ago I suggested a modular architecture that would let us compose flows with atomic operations. But it was archived and we should assume for now that we will keep current architecture of the library. 

We should favor features over infrastructure consistency as this library is in low maintenance but I believe we can easily modify the PR to use the existing  scrapers while supporting OTP with the great work of @orzarchi.

I verified it with Max account. You should verify the One zero as well.

Please checkout changes here #765. Feel free to pull it into your repo and create a PR from there if you prefer.

There are still the other code review comments but they are pretty much minor.

@orzarchi @erikash - wdyt?

## Adjustments done in the PR
1. use `ScraperCredentials` everywhere and adjust declaration to support otp
2. move relevant parts from `base-two-factor-auth-scraper` into `base-scraper`
3. delete `base-two-factor-auth-scrape`
4. extend One zero scraper with `BaseScraper`

# Motivation behind my suggestion
The new scraper `BaseTwoFactorAuthScraper` changes the current architecture of the library provided long time ago by @eshaham. I identify it in some places:
1. It doesn't emit the same progress events like the scrapers `BaseScraper` and `BaseScraperWithBrowser` exposed by method `onProgress`
5. Since one zero doesn't require puppeteer, the new scraper is similar to `BaseScraper` in the implementation. Once someone will need a an OTP scraper with puppeteer support we will need a fourth scraper for that.
6. It has some assumptions on the scraper credentials although once compiled it is not guaranteed  and might break at runtime

esakal:
@erikash thanks :)

> "Since one zero doesn't require puppeteer, the new scraper is similar to BaseScraper in the implementation. Once someone will need a an OTP scraper with puppeteer support we will need a fourth scraper for that". 

Today we have two base scrapers:
- `base-scraper` - used to scrape without puppeteer.
-  `base-scraper-with-browser` - extends base scraper with puppeteer scraping capabilities.

In the original PR @orzarchi introduced a new base scraper that didn't inherit from either. As one zero doesn't require puppeteer support, that scraper doesn't use puppeteer. My comment was that if we keep the third base scrapers it means that in future scrapers that require both OTP and puppeteer, we will need to provide a fourth base scraper. But in the suggested modification it is not an issue anymore. I hope it is more clear now :)
-

orzarchi:
hi @esakal - no problems, I'll fix @erikash's comments on top of your new PR with your changes.
Some thoughts about the discussion above:
1) I think we all proved once again that composition > inheritance, and that eventually the common functions in `base-scraper`/`base-scraper-with-browser` should be extracted to their own classes, and used by all concrete scrapers :)
2) I still feel like my approach with typing `ScraperCredentials` is better - if we define it per scraper, we can avoid jumping through hoops all over the codebase. Why? In the long term we can make the [scraper factory](https://github.com/orzarchi/israeli-bank-scrapers/blob/7475b6b4b7403c83670bf02e9bde4c1b2b4d6163/src/scrapers/factory.ts#L19) return the concrete scraper that accepts concrete credentials instead of the `Scraper` interface (using something like [this](https://stackoverflow.com/a/75391257/948829) technique), and remove files like [this](https://github.com/orzarchi/israeli-bank-scrapers/blob/7475b6b4b7403c83670bf02e9bde4c1b2b4d6163/src/definitions.ts#L25) as well as the code that tests it.
What do you feel about keeping the generic Credentials argument as I did [here](https://github.com/orzarchi/israeli-bank-scrapers/blob/8d30e5e1c6bb74456220da0d9ff3802c0c5d6cca/src/scrapers/one-zero.ts#L85-L95)?

esakal:
@orzarchi, I agree with you on both topics. Ideally, we should switch to a new composition-based architecture, but this will require active development work from developers to refactor all the scrapers. Unfortunately, this library is not currently being actively evolved by a group of developers; we are mostly focused on ongoing maintenance and limited by depending on developers that has actual access to each institute. This makes it challenging to change the architecture of the project.

However, if you do have a suggestion for a new architecture, testing it out on new scrapers could be a good way to gradually transition to it. For instance, One Zero might be a good candidate for this.

Regarding the credentials, you can certainly implement them. Just keep in mind that they will only work for those who initiate the scraper explicitly. Those who use the createScraper with Typescript might encounter runtime errors because the common credentials declaration is export declare type ScraperCredentials = Record<string, string>;

orzarchi:
Hey @erikash and @esakal, I've incorporated all of your suggestions and code.
I've implemented a compromise regrading the `ScraperCredentials` type - I've made each scraper define its own credentials, and made `ScraperCredentials` a big copied union of all of the different options.
This way we get more type safety everywhere at the cost of copied and pasted code.
In the future we can make the scraper factory function return concrete scraper types and get rid of `ScraperCredentials` altogether.

esakal (commented):
@orzarchi this is awesome! I want to merge it this weekend. There are few minor comments. Please review and decide if to fix them or not. I'm ok with having it like that. Please check my comment about the `...extraHeaders`.

Once you'll reply or commit fixes I'll merge it.

src/helpers/fetch.ts
```diff
@@ -30,20 +30,26 @@ export async function fetchGet<TResult>(url: string,
...
```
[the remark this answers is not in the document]
orzarchi: That isn’t strictly necessary, you can merge in empty/nulls with the triple dot :)

src/helpers/fetch.ts
```diff
@@ -30,21 +30,27 @@ export async function fetchGet<TResult>(url: string,
...
 
-export function fetchGetWithinPage<TResult>(page: Page, url: string): Promise<TResult | null> {
+export async function fetchGraphql<TResult>(url: string, query: string,
+  variables: Record<string, unknown> = {},
+  extraHeaders: Record<string, any> = {}): Promise<TResult> {
+  const result = await fetchPost(url, { operationName: null, query, variables }, extraHeaders);
+  if (result.errors?.length) {
+    throw new Error(result.errors[0].message);
+  }
+  return result.data as Promise<TResult>;
```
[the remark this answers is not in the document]
orzarchi: You’re right.
This is a bit of a “dirty” move, it just forces callers to pass in their expected return type in `<>` brackets instead of casting the result. I just think it’s slightly better than returning `any` like the rest of the methods in this file.
What do you think? I don’t mind removing the generic argument.

orzarchi:
> @orzarchi this is awesome! I want to merge it this weekend. There are few minor comments. Please review and decide if to fix them or not. I'm ok with having it like that. Please check my comment about the `...extraHeaders`.
> 
> Once you'll reply or commit fixes I'll merge it.

Sure, go ahead and merge it 🤩

esakal (approved):

esakal:
I approved the PR, it will be merged soon. Thanks @orzarchi  for your contribution

closes #765

eyaleng:
is there a plan to support other 2FA interfaces?
like Harel or Meitav for example

baruchiro:
> is there a plan to support other 2FA interfaces?
> like Harel or Meitav for example

I'm not sure I understand... Harel and Meitav are not Banks

eyaleng:
> > is there a plan to support other 2FA interfaces?
> > like Harel or Meitav for example
> 
> I'm not sure I understand... Harel and Meitav are not Banks

Well let's expand it. For all financial sources

baruchiro:
@eyaleng It will not be part of this project.

https://github.com/eshaham/israeli-bank-scrapers/issues/770
https://github.com/eshaham/israeli-bank-scrapers/issues/958
https://github.com/eshaham/israeli-bank-scrapers/issues/706

Issue #758, opened 2023-02-15
TypeScript upgrade


sagiba:
I am working on a private fork of the repo with some changes (will hopefully send a PR for some of them in the near future), and noticed that the TypeScript version that is used, 3.9.10, is very ancient. This is causing issues with bringing in some updated dependencies.

Any objection with trying to upgrade to a more modern version of TypeScript?

esakal:
Hi @sagiba sure, feel free to try upgrading. If it doesn't break anything we can merge it. If it does then please share what is broken or provide a draft PR to experiment with

PR #761, opened 2023-03-01
chore: upgrade typescript

chellem:
Main Upgrade:

`typescript` 3.9.10 => 4.9.5

and for compatibility purpose, upgrade of the following dependencies of:

- `jest`
- `ts-jest`
- `babel-jest`


esakal:
@chellem thank you for your contribution. I wonder if upgrading Typescript is considered as breaking change? What was the motivation for the upgrade? how did it impact you?

Two minor issues:
1. can you please remove the newly added yarn.lock file? the project is using npm (thus package-lock.json)
2. in some places you bypass Typescript by using `e as Error` or `rawTransaction as XYZ`. Is it needed? usually Typescript resolve the type for you and you don't need to do a bypass.

chellem:
Motivation: Feel that there was a need for an [upgrade](https://github.com/eshaham/israeli-bank-scrapers/issues/758) 

1. Remove `yarn.lock`
2. It's not a bypass, it's [type casting](https://stackoverflow.com/questions/54649465/how-to-do-try-catch-and-finally-statements-in-typescript).

esakal:
@chellem I see you closed the PR, do you want to re-open it or you closed it on purpose? 

The only question I have about the PR is about the usage of `e as Error` and `as undefined as string`. Beside that it looks good.

chellem:
@esakal do you have other suggestions that can be used instead of `e as Error` ?

PR #771, opened 2023-04-19
Fix release pipeline

daniel-hauser:
GitHub runners for `ubuntu-18.04` are not available, causing the workflow to hang with "Waiting for a runner to pick up this job..."

See hanging job here: https://github.com/eshaham/israeli-bank-scrapers/actions/runs/4746453978/jobs/8430124345

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


daniel-hauser:
@eshaham / @esakal Can you please approve?

esakal:
done. Thanks, it happened to me also in few other Github repositories, using `latest` is better then mentioning a specific version

PR #775, opened 2023-04-20
[visaCal] Use waitUntil to retry getting init data

daniel-hauser:
Sometimes the visaCal scraper fails with `"could not find 'init' data in session storage"`.
This is caused since the site might take more time to init.

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -251,7 +251,12 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials> 
   };
 
   async getCards() {
-    const initData = await getFromSessionStorage<InitResponse>(this.page, 'init');
+    const initData = await waitUntil(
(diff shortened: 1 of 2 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -251,7 +251,12 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
   };
 
   async getCards() {
-    const initData = await getFromSessionStorage<InitResponse>(this.page, 'init');
+    const initData = await waitUntil(
+      () => getFromSessionStorage<InitResponse>(this.page, 'init'),
+      'get init data in session storage',
+      10_000,
```
esakal, line 257: did you meant to have the _ here?

daniel-hauser: Yes, as a seperator for readability

erikash (approved):
Great PR :)

PR #780, opened 2023-04-22
fix(cal): Add transaction id and category back in the new visa site

itairaz1:


itairaz1:
Now we don't need extra requests for categories, since we get it in the first txs list request

itairaz1:
image
@erikash @orzarchi Awesome cooperate after 13~ years 😆

PR #787, opened 2023-05-13
hapoalim: change from reference number to transaction number

mnavon:
Up to now at the Hapoalim scraper, the transaction number was actually the institution code or the reference number.
To get the real transaction number you have to do another API call to get the extra details.
So in this PR, for every transaction, the scraper will get the extra details (using the `additionalTransactionInformation` flag).
If there is no transaction number, the scraper will use the institution code.


esakal:
@mnavon Thanks for your contribution. I merged your PR into the latest version.
Good night!
Eran

PR #795, opened 2023-07-01
Added credit utilization for credit cards (isracard-amex for now)

matan-elgabsi:
Not sure if this is on the roadmap, but needed to get my credit card utilization. added this quickly for isracard-amex.
Would love to hear your thoughts and correct as needed.

Thanks!

### DIFF
--- a/src/scrapers/base-isracard-amex.ts
+++ b/src/scrapers/base-isracard-amex.ts
@@ -102,6 +102,14 @@ interface ScrapedTransactionData {
   }>;
 }
 
+function getCreditUtilizationUrl(servicesUrl: string) {
+  return buildUrl(servicesUrl, {
+    queryParams: {
+      reqName: 'RikuzNetuneyCreditDigi',
+    },
+  });
+}
+
 async function fetchCreditUtilization(page: Page, servicesUrl: string): Promise<_.Dictionary<AccountCredit>> {
   const dataUrl = getCreditUtilizationUrl(servicesUrl);
   const dataResult = await fetchGetWithinPage<ScrapedAccountsWithinPageResponse>(page, dataUrl);
--- a/src/scrapers/base-isracard-amex.ts
+++ b/src/scrapers/base-isracard-amex.ts
@@ -54,7 +54,6 @@ interface ScrapedTransaction {
   paymentSumOutbound: number;
 }
 
-
 interface ScrapedAccount {
   index: number;
   accountNumber: string;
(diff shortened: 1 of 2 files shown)


esakal:
@matan-elgabsi  thanks for your contribution. sorry it took me time to respond.

I"m ok with the change. I have two thoughts to raise:

## Thought 1
the `TransactionAccount` is used also for banks. maybe it would be wise to separate it like

```diff
export interface TransactionsAccount {
  accountNumber: string;
+ credit: {
  creditUtilization?: number;
  creditTotal?: number;
+  } 
  balance?: number;
  txns: Transaction[];
}
```

## Thought 2
Do we want to always run the additional async call which if failed will fail the request and might impact response time? 
If we want to make it optional we can use the option https://github.com/eshaham/israeli-bank-scrapers/blob/2f9d73638b641c2c770f104a7887f7db12800cee/src/scrapers/interface.ts#L120C3-L120C35

I'm ok with both changes, just want to hear your thoughts.
@baruchiro  @erikash @matan-elgabsi  wdyt?

baruchiro:
Nice, why not.
Just to make sure, are you talking about מסגרת אשראי והניצול שלה?

Of course, @esakal said the right thoughts. Regarding 1: please set `credit?: ...` as optional to not break the code.
2 is important too.

matan-elgabsi:
Hi guys,
Thanks for the feedback!
Regarding 1:
Didn't notice that, sounds reasonable to separate them + set it as optional.
2:
I understand the problem, but not sure the flag you've mentioned is the right one, as the overhead here is just a single request.
I do think that adding some error handling and not failing the entire scraping (but instead just returning a null credit) is the right idea. What do you think?


FIY - I accidentally created the PR from a wrong account - After this discussion, I will create a new one instead.. sorry about that

esakal:
@matan-elgabsi sorry for the late response, my availability and access to the computer is small during Jule/Aug.

Regarding the second point, I understand it is only one call, I think the question is whether it is considered as mandatory date of that scraper or additional one. if additional then the flag fit to that purpose

esakal:
@matan-elgabsi hi, this is very old PR. What did you do on your end with it? is it still on your fork? 
Anyhow if it is relevant we can try to finalize and add it. 

@eshaham once approved we will need your help merging due to the stuck statuses tests

baruchiro:
The two thoughts from the first review still need to be done to finish this PR.

baruchiro:
Hi @matan-elgabsi, [Call for Code Owners](https://github.com/eshaham/israeli-bank-scrapers/issues/830)

matan-elgabsi:
Closed in favor of #857

PR #798, opened 2023-07-10
chore: upgrade to node v16

baruchiro:
I always get an error `Promise.any is not a function`. Sometimes I get it only in debugging mode in VSCode.

I suspect it is supported only in the latest v14 versions, and I think we should move to v16 anyway.


.github/workflows/nodeCI.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [14.x, 16.x]
```
esakal, line 11: can you remove 14.x? otherwise we will continue to check it against 14 in the CI

baruchiro: Sure

baruchiro: Done

eshaham:
@esakal done
@baruchiro would be helpful to update the Node version specified in the README.
It currently says Node 10 🤪 
Updating that would also retrigger the checks.

baruchiro:
@eshaham README is updated
Please note you need to change the required checks in order to make this PR able to pass.
image

eshaham:
I'll merge it forcefully, then.

PR #808, opened 2023-09-14
fix: Visa Cal - add missing transactions and fix the amount sign in case of credit (refund)

### LINKED ISSUE #807: visa-cal scraper incorrectly parsing credits
I use moneyman. When a transaction is a credit, I'm seeing the scraper return a negative `chargedAmount` and a positive `originalAmount`.

I expect them both to be positive in case of a credit (are there any cases where they should have a different sign?)

I'm seeing this only with visa-cal, the other scrapers seem fine.

itamarco:
Fixes #807 
This PR fix 2 issues with Visa Cal scraper:

1. In case of credits (refund) the "amount" field is still in minus instead of plus.
2. The scraper ignored the "immidiateDebits" field that may contains additional transactions

image

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -176,9 +176,15 @@ function createLoginFields(credentials: ScraperSpecificCredentials) {
 }
 
 function convertParsedDataToTransactions(parsedData: CardTransactionDetails[]): Transaction[] {
-  return parsedData
+  const bankAccounts = parsedData
     .flatMap((monthData) => monthData.result.bankAccounts)
+
+  const regularDebitDays = bankAccounts
     .flatMap((accounts) => accounts.debitDates)
+  const immediateDebitDays = bankAccounts
+    .flatMap((accounts) =>  accounts.immidiateDebits.debitDays)
+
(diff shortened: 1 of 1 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -193,11 +199,6 @@ function convertParsedDataToTransactions(parsedData: CardTransactionDetails[]):
       let chargedAmount = transaction.amtBeforeConvAndIndex * (-1);
       let originalAmount = transaction.trnAmt * (-1);
 
-      if (transaction.trnTypeCode === trnTypeCode.credit) {
-        chargedAmount = transaction.amtBeforeConvAndIndex;
-        originalAmount = transaction.trnAmt;
-      }
-
```
itamarco, line 200: Not needed anymore. sign is already correct for `chargedAmount`

esakal (approved):

esakal:
@itamarco  thanks for the contribution.
@eshaham can you check why the checks are on status waiting? I cannot merge the PR

esakal:
@eshaham I managed to merge a PR #811 but this one is still stuck on the checks. Please merge. hopefully new PRs will work also for me

PR #813, opened 2023-11-03
Support waitUntil parameter for puppeteer - Max scraper issue

roisec:
Hi 
I had an issue with Max scraper that stuck on scraping the login page in the last week.
When I did the debug, I saw that there was one request that got a time-out and stuck the scraping process on navigateTo function, which calls goto function of Puppeteer.
To fix that, there is a flag for Puppeteer when the  DOM document loaded content is finished - https://pptr.dev/api/puppeteer.puppeteerlifecycleevent, and when I used that it's fixes the issue.
So, I added support for waitUntil parameter from the scraper.

### DIFF
--- a/src/scrapers/max.ts
+++ b/src/scrapers/max.ts
@@ -291,7 +291,7 @@ class MaxScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials> {
       },
       postAction: async () => redirectOrDialog(this.page),
       possibleResults: getPossibleLoginResults(this.page),
-      waitUntil: 'domcontentloaded' as 'domcontentloaded',
+      waitUntil: 'domcontentloaded',
     };
   }
 
--- a/src/scrapers/max.ts
+++ b/src/scrapers/max.ts
@@ -291,6 +291,7 @@ class MaxScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials> {
       },
       postAction: async () => redirectOrDialog(this.page),
       possibleResults: getPossibleLoginResults(this.page),
+      waitUntil: 'domcontentloaded' as 'domcontentloaded',
     };
   }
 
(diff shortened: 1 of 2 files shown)


src/scrapers/max.ts
```diff
@@ -291,6 +291,7 @@ class MaxScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials> {
       },
       postAction: async () => redirectOrDialog(this.page),
       possibleResults: getPossibleLoginResults(this.page),
+      waitUntil: 'domcontentloaded' as 'domcontentloaded',
```
daniel-hauser, line 294: Consider changing the signature to:
```diff
+ getLoginOptions(credentials: ScraperSpecificCredentials): LoginOptions {
- getLoginOptions(credentials: ScraperSpecificCredentials) {
```
then you can do:
```suggestion
      waitUntil: 'domcontentloaded',
```

daniel-hauser:
@roisec Can you share the url of the slow resource?
if its a 3rd party script that is not used for the main flow we can block it like in #683 instead of waiting for it for >30s

roisec:
Hi @daniel-hauser ,
Thanks for your review!; I committed your changes.
We have one timeout request that stuck all the process - 
The problematic URL is:https://accessibility.max.co.il/loader/head
Regarding the waitUnitl parameter with "domcontentloaded," this is a permanent solution to solve further issues like this. I have the same problem in other places and need to support this flag for other scrapers.
BTW, I rarely have problems with Visacal that this flag should fix.

roisec:
@esakal @eshaham

doridembinsky:
Hey @daniel-hauser I got the same error on Max scraper and fixed it the same way locally..
is this scheduled to be merged soon?

daniel-hauser:
> Hey @daniel-hauser I got the same error on Max scraper and fixed it the same way locally.. is this scheduled to be merged soon?

I'm not a maintainer of israeli-bank-scrapers, but if you use [moneyman](https://github.com/daniel-hauser/moneyman) - this PR is included as a patch

eshaham (approved):
Sorry for taking so long to respond, this issue has been bugging me as well.
Thanks for the contribution @roisec 🙏

PR #819, opened 2023-12-15
[behatsdaa] Fix amount to be negative

daniel-hauser:
The current implementation of the scraper takes the amount from the `customerPrice` field returned from the behatsdaa api result,  which has the charged amount as a positive number.

This PR makes it negative as it's an expense

### DIFF
--- a/src/scrapers/behatsdaa.ts
+++ b/src/scrapers/behatsdaa.ts
@@ -33,7 +33,6 @@ type PurchaseHistoryResponse = {
 
 function variantToTransaction(variant: Variant): Transaction {
   const originalAmount = -variant.customerPrice;
-  
   return {
     type: TransactionTypes.Normal,
     identifier: variant.tTransactionID,
--- a/src/scrapers/behatsdaa.ts
+++ b/src/scrapers/behatsdaa.ts
@@ -33,6 +33,7 @@ type PurchaseHistoryResponse = {
 
 function variantToTransaction(variant: Variant): Transaction {
   const originalAmount = -variant.customerPrice;
+  
   return {
     type: TransactionTypes.Normal,
     identifier: variant.tTransactionID,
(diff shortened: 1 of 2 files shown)


daniel-hauser:
@esakal I believe the same issue might be in the **beyahad-bishvilha** scraper, but i have no account to verify.

src/scrapers/behatsdaa.ts
```diff
@@ -32,14 +32,15 @@ type PurchaseHistoryResponse = {
 };
 
 function variantToTransaction(variant: Variant): Transaction {
+  const originalAmount = -variant.customerPrice;
```
daniel-hauser, line 35:
```suggestion
  const originalAmount = -variant.customerPrice;
  
```

src/scrapers/behatsdaa.ts
```diff
@@ -32,14 +32,16 @@ type PurchaseHistoryResponse = {
...
+
```
daniel-hauser, line 36:
```suggestion
```

src/scrapers/behatsdaa.ts
daniel-hauser, line 35:
```suggestion
  // The price is positive, make it negative as it's an expense
  const originalAmount = -variant.customerPrice;
```

baruchiro (approved):

baruchiro:
Hi @daniel-hauser, [Call for Code Owners](https://github.com/eshaham/israeli-bank-scrapers/issues/830)

daniel-hauser:
@baruchiro - With the power granted to me in #830, i hereby declare this PR as tested 😉, can you merge it?

baruchiro:
Ha ha ha @daniel-hauser , the purpose is to get at least two testers for a change, to make sure your change is not over fitting for you. 😀

PR #822, opened 2023-12-31
Fix Mizrahi change password URL regex

baruchiro:
This pull request fixes the regular expression for the Mizrahi change password URL. The previous regex was incorrect and did not match the URL format. This PR updates the regex to correctly match the URL format.

### DIFF
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -47,7 +47,7 @@ const TRANSACTIONS_REQUEST_URLS = [
 ];
 const PENDING_TRANSACTIONS_PAGE = '/osh/legacy/legacy-Osh-p420';
 const PENDING_TRANSACTIONS_IFRAME = 'p420.aspx';
-const CHANGE_PASSWORD_URL = /https:\/\/www\.mizrahi-tefahot\.co\.il\/login\/\w+\/index\.html#\/change-pass/;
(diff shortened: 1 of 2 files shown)


src/scrapers/mizrahi.ts
```diff
@@ -70,9 +70,18 @@ function createLoginFields(credentials: ScraperSpecificCredentials) {
...
+  const oshXPath = `//a//span[contains(., "${checkingAccountTabHebrewName}") or contains(., "${checkingAccountTabEnglishName}")]`;
+  const oshTab = await options.page.$x(oshXPath);
+  return oshTab.length > 0;
+}
+
 function getPossibleLoginResults(page: Page): PossibleLoginResults {
   return {
-    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL, async () => !!(await page.$x(`//a//span[contains(., "${checkingAccountTabHebrewName}") or contains(., "${checkingAccountTabEnglishName}")]`))],
+    [LoginResults.Success]: [AFTER_LOGIN_BASE_URL, isLoggedIn],
     [LoginResults.InvalidPassword]: [async () => !!(await page.$(invalidPasswordSelector))],
```
baruchiro, line 85: Was always true because the return value of `page.$x` is an array, and `!!([])` is true.

PR #832, opened 2024-02-06
[Max] Add currency pocket support

daniel-hauser:
## Problem 
Max's new [currency pocket / ארנק מטח](https://www.max.co.il/cards/currency-pocket) feature added a new `planType`, causing the scraper to fail with the following message:
```
Unknown transaction type חיוב ארנק מטח
```

## Solution

This PR adds the missing plan type to `getTransactionType`, as well as a fallback using the `planTypeId` field (for future new plans).

### planTypeId
When looking on the transactions from my account we can see that plans 2 is `Installment`, and 5 is `Normal`
```js
Array.from(new Set(rawTransactions.map(t => [t.planName, t.planTypeId].join(':'))))
```
```json
[
    "רגילה:5",
    "חיוב עסקות מיידי:5",
    "תשלומים:2",
    "חיוב חודשי:5",
    "חיוב ארנק מטח:5"
]
```

Looking at the code in max, we can see that 2 and 3 are treated the same
```ts
<li *ngIf="
        (transaction.planTypeId === 2 || transaction.planTypeId === 3) &&
        transaction.dealData.amountLeft
    ">
    יתרה לתשלום:
    <span class="ltr-sum">
        {{
        transaction.dealData.amountLeft | maxCurrency: transaction.paymentCurrency
        }}
    </span>
</li>
```

## Other changes

With this change, i also:
- Moved all types to an Enum
- Renamed `txnType` to `planName` to align to the field in the max transaction object.
- `lint --fix` (causing the changes in visa-cal)

### DIFF
--- a/src/scrapers/max.ts
+++ b/src/scrapers/max.ts
@@ -23,6 +23,7 @@ interface ScrapedTransaction {
   originalCurrency: string;
   originalAmount: number;
   planName: string;
+  planTypeId: number;
   comments: string;
   merchantName: string;
   categoryId: number;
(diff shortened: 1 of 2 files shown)


baruchiro (approved):
Thank you!

baruchiro (approved):

PR #833, opened 2024-02-06
[Max] Add funds transfer support

daniel-hauser:
## Change

When sending/receiving money in apps like bit/PayBox, the sender/receiver's name (and the comment in PayBox) will not be included in the memo, but in 2 other fields: `fundsTransferComment` and `fundsTransferReceiverOrTransfer`.

This PR adds both fields to the memo.

### DIFF
--- a/src/scrapers/max.test.ts
+++ b/src/scrapers/max.test.ts
@@ -1,4 +1,4 @@
-import MaxScraper from './max';
+import MaxScraper, { getMemo } from './max';
 import {
   maybeTestCompanyAPI, extendAsyncTimeout, getTestsConfig, exportTransactions,
 } from '../tests/tests-utils';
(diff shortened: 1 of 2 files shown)


daniel-hauser:
This change is already running in moneyman (added in https://github.com/daniel-hauser/moneyman/pull/231)

src/scrapers/max.ts
```diff
@@ -165,6 +167,18 @@ function getChargedCurrency(currencyId: number | null) {
...
+  let memo = rawTransaction.comments;
+  if (rawTransaction.fundsTransferReceiverOrTransfer) {
+    memo += ` ${rawTransaction.fundsTransferReceiverOrTransfer}`;
+    if (rawTransaction.fundsTransferComment) {
+      memo += `: ${rawTransaction.fundsTransferComment}`;
+    }
+  }
+
+  return memo;
+}
```
baruchiro, line 180: What if the `rawTransaction.comments` is empty? This line will create a memo with a space as the first character:

```javascript
memo += ` ${rawTransaction.fundsTransferReceiverOrTransfer}`
```

And from your code I understand that the `rawTransaction.fundsTransferComment` comes always with `rawTransaction.fundsTransferReceiverOrTransfer`.

What do you think about adding a Unit Test for this function, with (as far as you can) real examples, so we will see exactly how the transaction is looks like and how the memo is?

daniel-hauser: Nice catch - comments are empty for fund transfers.
Although it will break existing moneyman hashes, i think it's the right thing to do and accept the risk.

Added some tests as well

src/scrapers/max.test.ts
```diff
@@ -50,3 +50,16 @@ describe('Max scraper', () => {
     exportTransactions(COMPANY_ID, result.accounts || []);
   });
 });
+
+describe.only('getMemo', () => {
```
daniel-hauser, line 54:
```suggestion
describe('getMemo', () => {
```

baruchiro (approved):
What an anoying logic!
And yet, you changed it to use consts without let 😍

baruchiro (approved):

Issue #830, opened 2024-02-05
Automatic Code Owners


baruchiro:
Hi everyone,

I just joined to be a maintainer in this lovely project 💖

Sometimes there are PRs here that wait for a long time because we can't test them without the relevant account.

I had an idea about a list of Code Owners- developers that accept to be tagged when there is a change in a bank they can help to test.

If you want to be notified and help us to test changes, please leave a message with the relevant bank accounts here.

Thank you!

| **Bank** | **Owners** |
| :---: | :---: |
| **Mizrahi** | @baruchiro  |
| **Cal** | @baruchiro @galbarm @daniel-hauser  |
| **Isracard** | @baruchiro @galbarm @daniel-hauser |
| **Max** | @baruchiro @galbarm @daniel-hauser @dimdimi4 |
| **Discount** | @galbarm |
| **Hapoalim** | @galbarm @daniel-hauser |
| **Amex** | @daniel-hauser |
| **Behatsdaa** | @daniel-hauser |
| **Yahav** | @gczobel |
| **Mercantile** | @EzzatQ @kfirarad |
| **Leumi** | @dimdimi4 |

**See the updated list on [`component_owners.yml`](https://github.com/eshaham/israeli-bank-scrapers/blob/master/.github/component_owners.yml)**

galbarm:
Hi, I can help testing Cal, Max, Isracard, Discount & Hapoalim

daniel-hauser:
@baruchiro I can help with Hapoalim, Cal, Max, Isracard, Amex and Behatsdaa

gczobel:
I can help with Bank Yahav.

baruchiro:
Missing code owners for
Beinleumi (@dudiventura?)
Leumi (@esakal?)
Beyahad-Bishvilha (@esakal?)
One-Zero (@orzarchi?)
Otzar-Hahayal (@matanelgabsi?)
Union-Bank

dimdimi4:
I can help with Leumi and Max.

nissant:
I can help with FIBI bank.

baruchiro:
I made a mistake in #836. The [dyladan/component-owners](https://github.com/dyladan/component-owners) require the accounts to be _collaborators_ to this project, but this is not our case.

We must create an automatic script to tag the relevant people on an issue/PR comment. (see #835)

matanelgabsi:
@baruchiro Write me up for Otsar hahayal :)
[...]

PR #838, opened 2024-03-01
feat: adding mercantile bank scraper

EzzatQ:
Changes:

* Created a new scraper for bank Mercantile based on the Discount bank. The implementations are identical as both mercantile and discount use the same system. The only difference between the two is the login url (`https://start.telebank.co.il/login/?bank=m`).  Code is based on [pull request](https://github.com/eshaham/israeli-bank-scrapers/pull/599) by @kfirarad.

* I have a mercantile account and I was able to test the scrapper successfully. Manually inspecting the payload returned by `scraper.scrape`:

```
{
  "success": true,
  "accounts": [
    {
      "accountNumber": "**redacted**",
      "balance":"**redacted**,
      "txns": [
        {
          "type": "normal",
          "identifier": "**redacted**",
          "date": "2024-02-01T22:00:00.000Z",
          "processedDate": "2024-01-30T22:00:00.000Z",
          "originalAmount": -20,
          "originalCurrency": "ILS",
          "chargedAmount": -20,
          "description": ""**redacted**",
          "status": "completed"
        },
        ...
      ]
    }
  ]
}
```


baruchiro:
Welcome @EzzatQ !!

I hope I will review it soon :-)

EzzatQ:
I continued testing this and it looks like we can only extract up to one year of transactions from current day (or actually 364 days to be exact), if requesting a longer period, mercantile's api will return an error message saying that "the requested information is not available, if you need more assistance call us on ....".

I'm wondering whether I should add some validation to the `options.startDate` param to avoid unexpected behavior. I can either throw an exception with a matching message, or take the maximum date between the passed `options.startDate` and `currentDate - 364`. What do you think?

baruchiro:
@EzzatQ see the Discount bank as an example of limiting the `startDate`. It is not the only scraper that's doing the same technique:

https://github.com/eshaham/israeli-bank-scrapers/blob/74479a1d9537039098d5f714471d6926da897b45/src/scrapers/discount.ts#L78-L80

baruchiro:
Also, I will add you to the Code Owners once we merge this, is that OK for you?
https://github.com/eshaham/israeli-bank-scrapers/issues/830

EzzatQ:
I am aware of the defaultStartMoment solution, in fact it is implemented in the mercantile scraper because it extends the discount scraper class. What I am talking about is validation of the user input, because if I supply a date that is before 1 year prior, I get an error from mercantile api, and that error is not propagated to the user. Even if it was, the error tells me nothing about why that issue happened aka the user won't know that the scraper is failing because of an invalid start date.

I personally have no problem to be added as a Code Owner. I don't know about @kfirarad as his PR is 3 years old

baruchiro:
Correct me if I'm wrong, but if you use the **max** date between the user `startDate` input and the _one year ago_ date, the start date will never go far beyond one year, so you will never get the error from the bank, right?

kfirarad:
Hi!
@EzzatQ Thanks for taking care of this one. I never had the chance to move it forward.
@baruchiro Feel free to add me as a code owner. Thank you for the suggestion.

baruchiro:
> Yes exactly

@EzzatQ So I don't understand when you will get the error, there will be no case for retrieving more than one year ago.

package-lock.json
```diff
@@ -1,14130 +1,8 @@
 {
   "name": "israeli-bank-scrapers",
   "version": "1.0.4",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
baruchiro, line 4: If you didn't changed the `package.json` the `package.lock.json` should not be changed too.

I guess it is because you're using the wrong version of NodeJS, which needs to be `v16.19.0`.

src/scrapers/mercantile.ts
```diff
@@ -0,0 +1,13 @@
...
+class MercantileScraper extends DiscountScraper {
+  getLoginOptions(credentials: ScraperSpecificCredentials) {
+    return {
+      ...super.getLoginOptions(credentials),
+      loginUrl: 'https://start.telebank.co.il/login/?bank=m',
+    };
+  }
+}
+
+export default MercantileScraper;
```
baruchiro, line 13: OMG I postponed this review for long time, just to finally review it and see it based on another scraper??

Sorry!!

baruchiro (approved):
Approved!

I see it is the same scraper like DiscountScraper. If it is working for you, I accept it.

baruchiro:
Status: I think it is a breaking change, I waiting for @eshaham @esakal

EzzatQ:
Sorry for disappearing, got very busy.
Thanks for reviewing and merging!

Issue #851, opened 2024-04-26
Behatsdaa scrapper return wrong sign


liranelisha:
all expenses made with the Behatsdaa card are returned with positive sign (ex. "230") while every other scrapper is returning a negative sign for expenses (ex. "-230").

all scrappers should return the same sign

baruchiro:
Do you plan to fix it? Sounds like a tiny change.

liranelisha:
do you mind setting my permissions to "write" or higher in order for me to be able to open a PR?

liranelisha:
i had to create a fork from the master as i do not have write permissions.
https://github.com/eshaham/israeli-bank-scrapers/pull/852 open, but i don't know if the workflows are on hold due to the fork or something else

baruchiro:
Yes, although forks are annoying, they are our way to contribute to open source.

You can see [how many forks I have](https://github.com/baruchiro?tab=repositories&q=&type=fork&language=&sort=) just because I wanted to patch something.

liranelisha:
it seems i had an older version where the sign was wrong

PR #852, opened 2024-04-29
Behatsdaa txns should be positive if expenses, same as the rest of the cards

liranelisha:
closes https://github.com/eshaham/israeli-bank-scrapers/issues/851

This pull request contains a small but significant change in the `src/scrapers/behatsdaa.ts` file. The `variantToTransaction` function has been modified to no longer negate the `customerPrice` when assigning it to `originalAmount`. This means that the transaction amount will now be represented as a positive value, rather than a negative one.

### DIFF
--- a/src/scrapers/behatsdaa.ts
+++ b/src/scrapers/behatsdaa.ts
@@ -32,8 +32,7 @@ type PurchaseHistoryResponse = {
 };
 
 function variantToTransaction(variant: Variant): Transaction {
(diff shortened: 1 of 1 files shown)


baruchiro:
Your change is wrong. As the comment said

> The price is positive, make it negative as it's an expense

As you can see, I debugged your change, and it looks like the price is indeed positive, and we need to negate it.

image


Even while I'm writing this message, I'm confused because issue #851 and this PR are saying the opposite.

PR #853, opened 2024-05-05
fix(beinleumi): added support for balance over 100K

arnonm:
Balances on BeinLeumi for over 100K are shown with number of commas. Changed replace to replaceAll, also added replace of NIS_CURRENCY_SYMBOL which appears if you have number of foreign accounts.


baruchiro:
Can you add a unit test with input examples and their expected output?

arnonm:
Sure. Do you have an example of a unit test format you are using? The
beinLeumi tests are really high level, and I don't see unit tests.

baruchiro:
See here:

https://github.com/eshaham/israeli-bank-scrapers/blob/master/src%2Fscrapers%2Fmax.test.ts#L54-L65

esakal:
@baruchiro I don't have this type of account but It looks legit from the code. do you want to wait for a unit test or to merge?

baruchiro (approved):

PR #856, opened 2024-06-08
fix(otsar-hahayal): updated success url

matanelgabsi:
New `rid` argument was added in the resulting query argument after login.
This fixes the issue, as the scraper is currently broken


baruchiro (approved):
Thank you!

ReaBar:
@matanelgabsi are you sure the regex shouldn't be at the end of the url to ignore the newly added rid path param? 
You can see an example in the [issue](https://github.com/eshaham/israeli-bank-scrapers/issues/858) I opened

baruchiro:
Please update your branch with the master branch

PR #863, opened 2024-06-17
fix visa cal login iframe

yanivfranco:
Visa cal has been broken for few days due to probably a url change for the login Iframe. 
Fixed by updating the iframe url.

@baruchiro  @daniel-hauser


baruchiro:
Is it solved by #865?

@erikash

baruchiro:
Solved by #865

PR #868, opened 2024-06-21
fix: login to visa-cal

esakal:
Updating the login flow to use the new url introduce in visa cal site


baruchiro (approved):
Did you run into a problem that led you to make this change?

esakal:
It seems to work again without changing the selector of the login

PR #879, opened 2024-08-16
fix(one-zero): transaction identifier

tomerroditi:
added an `identifier` field to the returned object from OneZero `fetchPortfolioMovements`


baruchiro (approved):
We don't have a reviewer for OneZero Bank (see #830).

Can I add you to the list?

tomerroditi:
Sure I can help with that. I'm not used to work with TS, I use python and now I'm slowly learning JS/TS for this project. will do my best :)

PR #880, opened 2024-08-18
feat: add bank pagi scraper

Avrhamwales:
Adding a scraper for Bank Pagi (a bank from the Bank Beinleumi Group).
Adding tests or Bank Pagi scraper.


baruchiro (approved):
Thank you!

@eshaham @esakal @erikash Please take a look to verify he is not missing something.

Regarding the version, do you think it introduces a breaking change for TypeScript consumers who built an object based on the Enum?

I think so, but I think it is not enough for a Major version 😉

baruchiro (approved):

Issue #881, opened 2024-09-01
Yahav scraper fails with Generic Error


gczobel:
Yahav scraper fails with Generic Error:
GENERIC Error: failed to find element matching selector ".dropdown-dir .selected-item-top"

Bank page was updated and the account ID elements is not the same.

PR #882, opened 2024-09-01
fix(yahav): account selector

gczobel:
fix: #881

Test:
```
 PASS  src/scrapers/yahav.test.ts (10.313 s)

Test Suites: 1 skipped, 17 passed, 17 of 18 total
Tests:       32 skipped, 22 passed, 54 total
Snapshots:   0 total
Time:        10.72 s
```

### DIFF
--- a/src/scrapers/yahav.ts
+++ b/src/scrapers/yahav.ts
@@ -61,15 +61,14 @@ function getPossibleLoginResults(page: Page): PossibleLoginResults {
 
 async function getAccountID(page: Page): Promise<string> {
   try {
-    // Attempt to extract the account ID using the selector
     const selectedSnifAccount = await page.$eval(ACCOUNT_ID_SELECTOR, (element: Element) => {
       return element.textContent as string;
     });
+    
     return selectedSnifAccount;
   } catch (error) {
-    // Handle the case where the selector is outdated or the element is not found
     const errorMessage = error instanceof Error ? error.message : String(error);
-    throw new Error(`Failed to retrieve account ID. Possible outdated selector. ${errorMessage}`);
+    throw new Error(`Failed to retrieve account ID. Possible outdated selector '${ACCOUNT_ID_SELECTOR}: ${errorMessage}`);
   }
 }
 
--- a/src/scrapers/yahav.ts
+++ b/src/scrapers/yahav.ts
@@ -23,7 +23,7 @@ const INVALID_DETAILS_SELECTOR = '.ui-dialog-buttons';
 const CHANGE_PASSWORD_OLD_PASS = 'input#ef_req_parameter_old_credential';
 const BASE_WELCOME_URL = `${BASE_URL}main/home`;
 
-const ACCOUNT_ID_SELECTOR = '.dropdown-dir .selected-item-top';
+const ACCOUNT_ID_SELECTOR = 'span.portfolio-value[ng-if="mainController.data.portfolioList.length === 1"]';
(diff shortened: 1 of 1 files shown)


src/scrapers/yahav.ts
```diff
@@ -60,10 +60,9 @@ function getPossibleLoginResults(page: Page): PossibleLoginResults {
 }
 
 async function getAccountID(page: Page) {
-  const selectedSnifAccount = await page.$eval(`${ACCOUNT_ID_SELECTOR}`, (option) => {
-    return (option as HTMLElement).innerText;
+  const selectedSnifAccount = await page.$eval(ACCOUNT_ID_SELECTOR, (e: Element) => {
+    return e.textContent ?? '';
```
baruchiro, line 64: If you return an empty string you didn't handle the unexpected behavior, you just hide it and it will be thrown to another place.

Please handle the possibility of an empty string where this function is called.

baruchiro (changes requested):

src/scrapers/yahav.ts
gczobel: @baruchiro I understand your point. The thing is that once you pass the login, there is no way the account ID is empty... except if the page was changed, that raises an error (and that's why I found this...)
WDYT? Raise another error here just in case?

gczobel: like this?

```
async function getAccountID(page: Page): Promise<string> {
  try {
    // Attempt to extract the account ID using the selector
    const selectedSnifAccount = await page.$eval(ACCOUNT_ID_SELECTOR, (element: Element) => {
      return element.textContent as string;
    });
    return selectedSnifAccount;
  } catch (error) {
    // Handle the case where the selector is outdated or the element is not found
    const errorMessage = error instanceof Error ? error.message : String(error);
    throw new Error(`Failed to retrieve account ID. Possible outdated selector. ${errorMessage}`);
  }
}
```

baruchiro: If you don't expect the string to be empty, ignore that case completely.

If the string is empty sometimes, the code will crash because of a page change.

You may be surprised, but this is the expected behavior- if the page was changed and something unexpected was received, then yes, crash it.

src/scrapers/yahav.ts
```diff
@@ -59,12 +59,18 @@ function getPossibleLoginResults(page: Page): PossibleLoginResults {
...
+    // Attempt to extract the account ID using the selector
+    const selectedSnifAccount = await page.$eval(ACCOUNT_ID_SELECTOR, (element: Element) => {
+      return element.textContent as string;
+    });
+    return selectedSnifAccount;
+  } catch (error) {
+    // Handle the case where the selector is outdated or the element is not found
+    const errorMessage = error instanceof Error ? error.message : String(error);
+    throw new Error(`Failed to retrieve account ID. Possible outdated selector. ${errorMessage}`);
+  }
```
baruchiro, line 73:
```suggestion
  try {
    const selectedSnifAccount = await page.$eval(ACCOUNT_ID_SELECTOR, (element: Element) => {
      return element.textContent as string;
    });
    
    return selectedSnifAccount;
  } catch (error) {
    const errorMessage = error instanceof Error ? error.message : String(error);
    throw new Error(`Failed to retrieve account ID. Possible outdated selector '${ACCOUNT_ID_SELECTOR}: ${errorMessage}`);
  }
```

baruchiro (approved):

PR #884, opened 2024-10-02
feat: allow to reuse a browser by passing a `browserContext`

daniel-hauser:
**Motivation:**
- Enables efficient scraping of multiple accounts in parallel by allowing to reuse a browser.
- Enables parallel runs of the same scraper with different users by creating a browser context for each user.

**Changes:**
- Modified the page creation logic to allow passing a `browserContext` instead of a `browser`.
- The `ScraperOptions` interface now has a union of three `BrowserOptions` interfaces:
  - `DefaultBrowserOptions` for when the scraper creates the browser.
  - `ExternalBrowserOptions` for when the caller supplies an external single-use browser.
      - An optional `skipCloseBrowser` flag was added to allow the caller to manage the browser lifecycle.
  - `ExternalBrowserContextOptions` for when the caller supplies a `BrowserContext` and manages the browser lifecycle.


src/scrapers/interface.ts
baruchiro, line 1: Why did you changed the order of the properties?

It is very confusing to review it now, and as you can see, we don't have much time to review it, I'm doing it from my phone.

daniel-hauser: The GitHub diff is indeed bad for this change.

The reason properties moved around is that i extracted all options related to the browser from the existing `ScraperOptions` type to a new `ScraperBrowserOptions` type

src/scrapers/interface.ts
```diff
@@ -26,7 +26,7 @@ export interface FutureDebit {
   bankAccountNumber?: string;
 }
 
-export interface ScraperOptions {
+export type ScraperOptions = ScraperBrowserOptions & {
```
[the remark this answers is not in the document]
daniel-hauser: `ScraperBrowserOptions` is a union type, Interfaces can only extend object types with statically known members

src/scrapers/base-isracard-amex.ts
```diff
@@ -204,9 +204,9 @@ function convertTransactions(txns: ScrapedTransaction[], processedDate: string):
   });
 }
 
-async function fetchTransactions(page: Page, options: ExtendedScraperOptions, startMoment: Moment, monthMoment: Moment): Promise<ScrapedAccountsWithIndex> {
-  const accounts = await fetchAccounts(page, options.servicesUrl, monthMoment);
-  const dataUrl = getTransactionsUrl(options.servicesUrl, monthMoment);
+async function fetchTransactions(page: Page, options: ScraperOptions, companyServiceOptions:CompanyServiceOptions, startMoment: Moment, monthMoment: Moment): Promise<ScrapedAccountsWithIndex> {
+  const accounts = await fetchAccounts(page, companyServiceOptions.servicesUrl, monthMoment);
+  const dataUrl = getTransactionsUrl(companyServiceOptions.servicesUrl, monthMoment);
```
baruchiro, line 209: Do you have to split the `companyServiceOptions` from `options`?

src/scrapers/base-scraper-with-browser.ts
```diff
@@ -169,6 +131,57 @@ class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends Ba
...
+    debug('initialize browser page');
+    if ('browserContext' in this.options) {
+      debug('Using the browser context provided in options');
+      return this.options.browserContext.newPage();
+    }
+
+    if ('browser' in this.options) {
+      debug('Using the browser instance provided in options');
+      const { browser } = this.options;
+
```
baruchiro, line 144: I need to review it, but I don't have a time for it

src/scrapers/base-isracard-amex.ts
daniel-hauser: These fields are not `ScraperOptions` but amex/isracard specific values that are internal to this file - defined inside `fetchData` and merged with the scraper options.

src/scrapers/base-scraper-with-browser.ts
daniel-hauser: @baruchiro, I aggrege that reviewing large changes in GH mobile is hard.
Attached here is the `o1` answer for the changes, LMK if it makes the review easyer.
__________

_AI Generated:_

<details>
<summary>
Longe explanation:
</summary>

### **Changes Introduced:**

- **Added Support for External `BrowserContext`:**
  - **New Logic Added:**
    - In the `initializePage()` method, we now first check if a `browserContext` is provided in `this.options`.
    - **If a `browserContext` is provided:**
      - We create a new page using `browserContext.newPage()`.
      - This allows users to supply their own `BrowserContext`, offering more flexibility when the browser context is managed externally.

---

### **Logic Unchanged:**

- **Existing Browser Initialization Logic Remains the Same:**
  - **If no `browserContext` is provided:**
    - We proceed to check if a `browser` is provided in `this.options` (same as before).
      - **If a `browser` is provided:**
        - We log a debug message: `'Using the browser instance provided in options'`.
        - We create a new page using `browser.newPage()`.
        - For backward compatibility, if `skipCloseBrowser` is not set to `true`, we add a cleanup function to close the browser after scraping.
      - **If no `browser` is provided:**
        - We launch a new browser instance as before, using options like `executablePath`, `args`, `showBrowser`, and `timeout` from `this.options`.
        - We add a cleanup function to close the browser after scraping.
        - If `prepareBrowser` is provided, we call it with the new browser instance.

- **Resource Cleanup Logic Remains the Same:**
  - We maintain a `cleanups` array to manage cleanup functions.
  - During termination (`terminate()` method), we execute all cleanup functions in reverse order to ensure proper resource disposal.
  - This includes closing pages and browsers as needed.
  
- **Page Creation and Return:**
  - After handling the browser or browser context, we create a new page and return it.
  - This part of the logic is unchanged.

</details>

### **Summary:**

- **What's New:**
  - Added support for an external `BrowserContext` in `initializePage()`, allowing more flexibility in browser management.

- **What's Unchanged:**
  - The existing logic for:
    - Handling an external `Browser` instance.
    - Launching a new browser if none is provided.
    - Preparing the browser with `prepareBrowser` if provided.
    - Managing resource cleanup with the `cleanups` array.
    - Creating and returning a new page for the scraper to use.
  - All existing behavior remains intact, ensuring backward compatibility.

daniel-hauser: @baruchiro - Will appreciate your approval when you have time

daniel-hauser: FYI, the new code is successfully running twice a day since i added https://github.com/daniel-hauser/moneyman/pull/356  on Oct 5

src/scrapers/base-scraper-with-browser.ts
```diff
@@ -169,6 +131,57 @@ class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends Ba
     });
   }
 
+  private async initializePage() {
+    debug('initialize browser page');
+    if ('browserContext' in this.options) {
+      debug('Using the browser context provided in options');
+      return this.options.browserContext.newPage();
+    }
```
baruchiro, line 139: The common way to do this condition is `if (this.options.browserContext)`

src/scrapers/base-scraper-with-browser.ts
```diff
@@ -169,6 +131,57 @@ class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends Ba
...
+
+    if ('browser' in this.options) {
+      debug('Using the browser instance provided in options');
+      const { browser } = this.options;
```
baruchiro, line 143: same here

src/scrapers/base-scraper-with-browser.ts
```diff
@@ -169,6 +131,57 @@ class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends Ba
...
+
+      /**
+       * For backward compatibility, we will close the browser even if we didn't create it
+       */
+      if (!this.options.skipCloseBrowser) {
+        this.cleanups.push(async () => {
+          debug('closing the browser');
+          await browser.close();
+        });
```
baruchiro, line 152: If the browser exists and you run multiple accounts, this cleanup task will be pushed for each account (except the first one):
1. It will run the same cleanup multiple times
1. If only one account triggered, is it will close the browser?

src/scrapers/base-scraper-with-browser.ts
daniel-hauser: Since `ScraperBrowserOptions` is now a union type and `this.options.browserContext` will emit the following error:
```
Property 'browserContext' does not exist on type 'ScraperBrowserOptions'.
  Property 'browserContext' does not exist on type 'ExternalBrowserOptions'.
```

Only after the `'browserContext' in this.options` check, TS can narrow the type of `this.options` to `ExternalBrowserContextOptions`

src/scrapers/base-scraper-with-browser.ts
daniel-hauser: Same as in https://github.com/eshaham/israeli-bank-scrapers/pull/884#discussion_r1921944357 - but with `ExternalBrowserOptions`

src/scrapers/base-scraper-with-browser.ts
daniel-hauser: The logic here will run from `scrape > initialize` for each scrape, and we will have one cleanup call.

This is the exact same behavior as the current [`terminate` implementation](https://github.com/eshaham/israeli-bank-scrapers/blob/26837defff377f336b246117f715a20923e8c12a/src/scrapers/base-scraper-with-browser.ts#L277C2-L281C32) that is:

```
    if (!this.browser) {
      return;
    }

    await this.browser.close();
```

src/scrapers/base-scraper-with-browser.ts
daniel-hauser: Yes, the diff is large since i separated the flow and types to the 2 existing variations (with and without a browser) and then added one more.

I could have not separate, but having separate types bakes the code and types better (with the price being the diff 😅)

src/scrapers/interface.ts
```diff
@@ -26,49 +26,40 @@ export interface FutureDebit {
   bankAccountNumber?: string;
 }
 
-export interface ScraperOptions {
+interface ExternalBrowserOptions {
```
baruchiro, line 29: Please explain how one `ScraperOptions` interface became to be a union of **three** interfaces with another fourth interface.

Sorry, something is not clear here, as you can see.

Maybe it is naming only, but it is not clear to me what is the difference between `ExternalBrowserOptions` and `ExternalBrowserContextOptions`, do are they both new ways to reuse a browser?
And if they are, why is there a third interface `DefaultBrowserOptions`? I guess this is the regular way of launching a new browser for each scraper.

Now after your change it is clear there are two different things inside one interface: Browser Settings and Scraping Settings. Maybe we need to split the interfaces or at least insert one of the subjects inside a property (`scrapeConfig: ScrapeConfig`) in the other one.

daniel-hauser: > Please explain how one ScraperOptions interface became to be a union of three interfaces with another fourth interface.
...
Maybe it is naming only, but it is not clear to me what is the difference between ExternalBrowserOptions and ExternalBrowserContextOptions, do are they both new ways to reuse a browser?

The code had two ways to create a page: 
1. The default way - the scraper creates the browser, the caller can supply some options for the creation.
    - Named `DefaultBrowserOptions` in this PR
2. The non-default way - the caller supplies an external **single use** browser (because `browser.close()` is called)
     - Named `ExternalBrowserOptions` in this PR

In this PR i added the third way, where the caller only supplies a [BrowserContext](https://pptr.dev/api/puppeteer.browsercontext) and is responsible for the browser creation and teardown (named `ExternalBrowserContextOptions`)

> Now after your change it is clear there are two different things inside one interface: Browser Settings and Scraping Settings. Maybe we need to split the interfaces or at least insert one of the subjects inside a property (scrapeConfig: ScrapeConfig) in the other one.

In order to stay backwards compatible, I didn't want to add a new `browserOptions` property to `ScraperOptions`, therefore left it as `export type ScraperOptions = ScraperBrowserOptions & { ... /* All non-browser options */ }`

baruchiro: Here is the current implementation on `base-scraper-with-browser`:
```typescript
    if (typeof this.options.browser !== 'undefined' && this.options.browser !== null) {
      debug('use custom browser instance provided in options');
      this.browser = this.options.browser;
    } else {
      const executablePath = this.options.executablePath || undefined;
      const args = this.options.args || [];
      const { timeout } = this.options;

      const headless = !this.options.showBrowser;
      debug(`launch a browser with headless mode = ${headless}`);
      this.browser = await puppeteer.launch({
        env,
        headless,
        executablePath,
        args,
        timeout,
      });
    }
```

It seems you can already inject your browser instance and control its launch and teardown. Why do you need the `BrowserContext`?

daniel-hauser: My motivation for this change is to scrape multiple accounts in parallel. The current implementation has the following issues:
1. Creating a browser for each scraper is not efficient.
2. Even if I somehow prevent the scraper from closing the browser, since the pages are created in the default context, running the same scraper with different users at the same time will not work.

Passing a `browserContext` solves both issues

baruchiro: OK then.

Can you please create a simple example for each interface to explain the differences and the capabilities?

I want us to add these examples, I'm not sure if put them as JSDoc on the interfaces or inside the README (currently the README directs to the interfaces by "Check the options declaration [here](./src/scrapers/interface.ts#L29) for available options.")

daniel-hauser: >  inside the README

Done

README.md
```diff
@@ -129,6 +129,45 @@ The return value is a list of scraper metadata:
...
+  companyId: CompanyTypes.leumi,
+  startDate: new Date('2020-05-01'),
+  browserContext
+};
+const scraper = createScraper(options);
+const scrapeResult = await scraper.scrape({ username: 'vr29485', password: 'sometingsomething' });
+await browserContext.close();
+await browser.close();
+// TODO: Use scrapeResult
+```
```
baruchiro, line 170: image

It is not clear why I want to use browserContext instead of browser (except the incognito usecase, which if it is the only use case, I might solve the problem in the other way).

I remember you said that browserContext allows you to work with multiple tabs. Can you add this use case to the example?

README.md
```diff
@@ -130,7 +130,56 @@ The return value is a list of scraper metadata:
...
+This option allows you to provide an externally managed browser context. This is useful when you want to manage the browser context yourself.
+
+Example:
+
+```typescript
+import puppeteer from 'puppeteer';
+import { CompanyTypes, createScraper } from 'israeli-bank-scrapers';
+
+const browser = await puppeteer.launch();
+const browserContext = await browser.createIncognitoBrowserContext();
```
daniel-hauser, line 169:
```suggestion
const browserContext = await browser.createBrowserContext();
```

README.md
```diff
@@ -130,7 +130,56 @@ The return value is a list of scraper metadata:
...
+const options = {
+  companyId: CompanyTypes.leumi,
+  startDate: new Date('2020-05-01'),
+  browserContext
+};
+const scraper = createScraper(options);
+const scrapeResult = await scraper.scrape({ username: 'vr29485', password: 'sometingsomething' });
+await browserContext.close();
```
daniel-hauser, line 177:
```suggestion
```

README.md
```diff
@@ -130,7 +130,56 @@ The return value is a list of scraper metadata:
...
+  skipCloseBrowser: true, // Or false [default] if you want it to auto-close
+};
+const scraper = createScraper(options);
+const scrapeResult = await scraper.scrape({ username: 'vr29485', password: 'sometingsomething' });
+await browser.close(); // Or not if `skipCloseBrowser` is false
+```
+
+### ExternalBrowserContextOptions
+
+This option allows you to provide an externally managed browser context. This is useful when you want to manage the browser context yourself.
```
daniel-hauser, line 160:
```suggestion
This option allows you to provide a [browser context](https://pptr.dev/api/puppeteer.browsercontext). This is useful if you don't want to share cookies with other scrapers (i.e. multiple parallel runs of the same scraper with different users) without creating a new browser for each scraper.
```

README.md
daniel-hauser: @baruchiro I updated the readme, hope it's clearer now.

> except the incognito usecase, which if it is the only use case, I might solve the problem in the other way)

The way to solve is to create a browser context, launching the browser itself in incognito will have that same problem as cookies will still be shared in the lifetime of the browser)

> I remember you said that browserContext allows you to work with multiple tabs. Can you add this use case to the example?

The `browserContext` is not for working with multiple tabs, it's for being able to login with multiple users on the same website withput creating multiple browsers - [from the docs](https://pptr.dev/api/puppeteer.browsercontext#:~:text=Each%20context%20has%20isolated%20storage%20(cookies/localStorage/etc.)):
> Each context has isolated storage (cookies/localStorage/etc.)

baruchiro (approved):

daniel-hauser:
@baruchiro Thanks for approving! Do you have an estimation for the merging of the PR?

PR #893, opened 2025-01-12
fix(beinleumi): account with no balance (inactive)

### LINKED ISSUE #892: Beinleumi: failed to find element matching selector ".main_balance"
Basically the same as #610 - only the first page of transactions has the `main_balance`, but following pages are missing it. This is causing the following `GENERIC` error:
```
Error: failed to find element matching selector ".main_balance"
```
Looks like it was fixed in #611 back in 2021, but I'm facing it now in 2025 :/

I'm using v5.2.2 of `israeli-bank-scrapers`.
Also facing this issue when using [Caspion](https://github.com/brafdlog/caspion) v2.6.0.

# Update
The issue is actually due to old inactive accounts not having the `main_balance` element. See https://github.com/eshaham/israeli-bank-scrapers/issues/892#issuecomment-2585740536

whatuserever:
In Beinleumi, inactive accounts expectedly do no have a balance. See the comparison below between an active and inactive accounts.

This PR fixes #892  by making the `balance` optional for Beinleumi

### Active
image

### Inactive
image

## Test
Ran the scraper and printed `scrapeResult.accounts`
### Before
```
Error: failed to find element matching selector ".main_balance"
```
### After
```
[
  {
    accountNumber: '1',
    txns: [ [Object], [Object], [Object], [Object], [Object] ],
    balance: 12345.67
  },
  { accountNumber: '2', txns: [], balance: undefined }
]
```


src/scrapers/base-beinleumi-group.ts
```diff
@@ -280,7 +284,7 @@ async function fetchAccountData(page: Page, startDate: Moment) {
   return {
     accountNumber,
     txns,
-    balance,
+    balance: balance === null ? undefined : balance,
```
whatuserever, line 287: The linter didn't like `balance` being `null`, likely because of the way `TransactionsAccount` is defined:
https://github.com/eshaham/israeli-bank-scrapers/blob/28fd0e943e2996b7be7f1f8828d324b57e2846b4/src/transactions.ts#L3

A quick search suggested replacing it with `undefined`. Let me know if there's a more elegant approach.

whatuserever: Now that I think about it, I can probably just return `undefined` directly from `getCurrentBalance`, right? Let me know what you prefer 🤷‍♂️

whatuserever: Changed it to just return `undefined` in the first place :)

whatuserever:
@baruchiro could you please review :pray:

src/scrapers/base-beinleumi-group.ts
```diff
@@ -258,6 +258,10 @@ async function getAccountTransactions(page: Page) {
 }
 
 async function getCurrentBalance(page: Page) {
+  const balanceElement = await page.$(CURRENT_BALANCE);
+  if (!balanceElement) {
+    return undefined;
+  }
   const balanceStr = await page.$eval(CURRENT_BALANCE, (option) => {
     return (option as HTMLElement).innerText;
   });
```
baruchiro, line 267: I don't know your JS Scraping experience, but do you think once you have a reference to the `balanceElement`, get its `innerText` (instead of using `eval` again on the same selector)?

whatuserever: I have no experience at all with either JS or scraping :upside_down_face: 

I looked at the [Puppeteer docs](https://pptr.dev/api/puppeteer.page.evaluate#example-3) now, and it looks like we can do `balanceElement.evaluate`. This is still another evaluation, but I guess it's slightly more efficient as it shouldn't be searching for the element in the entire page? Not sure tbh.

baruchiro: You see, it is not so important, and I don't think efficiency is critical here.

But it is less confusing, and the code is more straightforward—find an element, and if it exists, take its innerText.

baruchiro (approved):

baruchiro:
@whatuserever Can I register you as a tester for Beinleumi? #830

Issue #894, opened 2025-01-19
Can't scrape more than one year's worth of data


whatuserever:
I'm trying to scrape all available data from my financial accounts. However, when I set a start date for more than one year ago, it gets overwritten with a date that is about one year ago from today.

Looking at the code, this seems to be the root cause:
https://github.com/eshaham/israeli-bank-scrapers/blob/26837defff377f336b246117f715a20923e8c12a/src/scrapers/base-beinleumi-group.ts#L343-L345

I noticed that this logic [exists in most (all?) scrapers](https://github.com/search?q=repo%3Aeshaham%2Fisraeli-bank-scrapers%20moment.max&type=code).

Testing Beinleumi and Max without this, i.e. with the following diff, seems to work perfectly fine:
```diff
-    const startMoment = moment.max(defaultStartMoment, moment(startDate));
+    const startMoment = moment(startDate);
```

Is there a reason for this limitation? Can it be removed?

baruchiro:
I think it should be defined like this only for websites that disallow fetching more than one year before.

whatuserever:
I figured the history limit for Beinleumi and Max and created PRs. Hopefully others can do the same for the rest :)

whatuserever:
I see #895 and #896 were approved, but I don't have permissions to merge. @baruchiro could you please merge theme as well? 🙏

baruchiro:
I'm waiting a while to give others a chance to review.

PR #896, opened 2025-01-19
fix(beinleumi): allow scraping older transactions

whatuserever:
Beinleumi allows retrieving data from 1600 :joy:
image

Any earlier than that and it fails:
image

Mitigates https://github.com/eshaham/israeli-bank-scrapers/issues/894 for Beinleumi

### Test
Set `startDate` in `.tests-config.js` to 5 years ago and ran tests:
```
 PASS  src/scrapers/beinleumi.test.ts (35.25 s)
```
`beinleumi.csv` output file had transactions going all the way back

### DIFF
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -341,8 +341,9 @@ class BeinleumiGroupBaseScraper extends BaseScraperWithBrowser<ScraperSpecificCr
 
   async fetchData() {
     const defaultStartMoment = moment().subtract(1, 'years').add(1, 'day');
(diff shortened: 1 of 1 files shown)


baruchiro:
Let me understand it... If you set this earlier date, is it really returning your five-years-old transactions?

I want to be sure it actually retrieves the data.

whatuserever:
My account is about 3 years old, so I can't test too far back (definitely not to 1600 😂).

But yes, it does return all of my account's data, and there's no issue with specifying a date before the account's creation.

baruchiro (approved):

PR #899, opened 2025-01-31
refactor: remove old banks keys **BREAKING CHANGES**

baruchiro:
BREAKING CHANGES: the `hapoalimBeOnline` was replaced with `hapoalim` and the `leumiCard` was replaced with `max`. All the new keys are already there for months.


esakal:
hi. I don't have those institutes. I understand they don't exists any more? if so lets merge it

Nic3Guy:
i have them both, i can test it today later the evening, i will update it thats what you're waiting for

baruchiro:
@esakal They are just old keys that replcaed with new keys, if you will look closely on the PR diff, you will se they are directed into scrapers that are also directed by new keys:

image

Nic3Guy:
If this helps I just test PR with my creds for `max` and `hapoalim` and it works ✅ 
image

PR #903, opened 2025-03-01
fix(beinleumi): support new UI

### LINKED ISSUE #901: Beinleumi getting timeout after login - probably due to the new UI
When trying to scrape "BeinLeumi" account, after the login (which is successful) it pauses and get timeout with out scraping

Waiting for the debugger to disconnect...

### Additional info maybe it will help:
1. There is an overlay that suggest walkthough of the new UI - but even if manually cancelling it doesn't help, also it always return even if you tell it not to.
2. I wasn't able to locate the `#matafLogoutLink` when I search with devTools

### Screenshot
the overlay:
Image

Nic3Guy:
Fix #901 

--------------

The BeinLeumi have new UI, the transactions page is wrap with iFrame now
needed to adjust the `base-beinleumi-group.ts` 

I don't know if I did it with the best practices of Node.js, but it works, and i also ran the test and all pass (see screenshot)
Screenshot 2025-03-01 at 21 57 20

### DIFF
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -134,7 +134,7 @@ function extractTransactionDetails(txnRow: TransactionsTr, transactionStatus: Tr
   return item;
 }
 
-async function getTransactionsColsTypeClasses(page: Page, tableLocator: string): Promise<TransactionsColsTypes> {
+async function getTransactionsColsTypeClasses(page: Page | Frame, tableLocator: string): Promise<TransactionsColsTypes> {
   const result: TransactionsColsTypes = {};
   const typeClassesObjs = await pageEvalAll(page, `${tableLocator} tbody tr:first-of-type td`, null, (tds) => {
     return tds.map((td, index) => ({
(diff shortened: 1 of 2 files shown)


Nic3Guy (commented):
Fixing the lint

baruchiro:
Just one question before reviewing your change, did you ran the tests with your username and password?

src/scrapers/base-beinleumi-group.ts
```diff
@@ -298,11 +298,24 @@ async function getAccountIdsBySelector(page: Page): Promise<string[]> {
...
 }
 
+async function getTransactionsFrame(page: Page): Promise<Frame> {
+  await sleep(5000);
+  const frames = page.frames();
+  const targetFrame = frames.find(f => f.name() === 'iframe-old-pages' || f.url().includes('some-part-of-url'));
+
+  if (!targetFrame) {
+    throw new Error('Target iframe not found');
+  }
```
baruchiro, line 308: Maybe more detailed and less technical message?

`iframe name='iframe-old-pages' with URL includes 'some-part-of-url' was not found on the page to login with`

Nic3Guy: Changed

Nic3Guy:
> Just one question before reviewing your change, did you ran the tests with your username and password?

Yes, but only for "beinleumi" I didn't ran any other scraper because I didn't modified any, is that ok?

arieltobiana:
@baruchiro—It would be great if you could insert this PR into a new release since the Benleuim scrapper has been broken for three weeks already.
Thanks!

baruchiro:
@arieltobiana did you tested this code with your credentials too?

Sorry for the delay 😨

arieltobiana:
@baruchiro - I am in military reserve without the ability to test it
I suggest creating a new release with it. The scrapper is already broke, so it can't be worse :)

PR #908, opened 2025-03-15
fix(beinleumi): support both new and old UI

erikash:
This pull request includes several updates to the `src/scrapers/base-beinleumi-group.ts` file to enhance compatibility with both the new and old user interfaces of the banking website. The changes focus on updating the selectors and improving the resilience of iframe handling.

Key updates include:

### UI Compatibility Enhancements:
* Updated `getPossibleLoginResults` function to recognize patterns from both the new and old UI.
* Enhanced `waitForPostLogin` function to wait for elements specific to both UI versions.

### Iframe Handling Improvements:
* Introduced a retry mechanism in `getTransactionsFrame` to handle cases where the iframe might not be immediately available.
* Defined a constant `IFRAME_NAME` for the iframe name to avoid hardcoding it multiple times.

### Account Data Fetching:
* Modified the logic in `fetchAccounts` to handle both single and multiple account scenarios more robustly, and to use the appropriate page or iframe based on availability.

### DIFF
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -319,26 +319,33 @@ async function getTransactionsFrame(page: Page): Promise<Frame | null> {
   return null;
 }
 
-async function fetchAccounts(page: Page, startDate: Moment) {
-  const accounts: TransactionsAccount[] = [];
+async function selectAccount(page: Page, accountId: string) {
+  await page.select('#account_num_select', accountId);
+  await waitUntilElementFound(page, '#account_num_select', true);
+}
+
+async function fetchAccountDataBothUIs(page: Page, startDate: Moment) {
+  // Try to get the iframe for the new UI
+  const frame = await getTransactionsFrame(page);
+    
+  // Use the frame if available (new UI), otherwise use the page directly (old UI)
+  const targetPage = frame || page;
+  return fetchAccountData(targetPage, startDate);
+}
+
+async function fetchAccounts(page: Page, startDate: Moment):Promise<TransactionsAccount[]> {
   const accountsIds = await getAccountIdsBySelector(page);
-  
-  // Convert single account case to an array with one item
-  const accountsToProcess = accountsIds.length <= 1 ? [''] : accountsIds;
-  
-  for (const accountId of accountsToProcess) {
-    // Only select account if we have multiple accounts
-    if (accountsIds.length > 1) {
-      await page.select('#account_num_select', accountId);
-      await waitUntilElementFound(page, '#account_num_select', true);
-    }
+
+  if (accountsIds.length <= 1) {
+    const accountData = await fetchAccountDataBothUIs(page, startDate);
+    return [accountData];
+  }
+
+  const accounts: TransactionsAccount[] = [];
+  for (const accountId of accountsIds) {
+    await selectAccount(page, accountId);
     
-    // Try to get the iframe for the new UI
-    const frame = await getTransactionsFrame(page);
-    
-    // Use the frame if available (new UI), otherwise use the page directly (old UI)
-    const targetPage = frame || page;
-    const accountData = await fetchAccountData(targetPage, startDate);
+    const accountData = await fetchAccountDataBothUIs(page, startDate);
     accounts.push(accountData);
   }
   
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -54,7 +54,7 @@ export function getPossibleLoginResults(): PossibleLoginResults {
   const urls: PossibleLoginResults = {};
   urls[LoginResults.Success] = [
     /fibi.*accountSummary/,  // New UI pattern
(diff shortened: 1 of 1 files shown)


Nic3Guy:
@erikash please don't merge this yet, I test it (I'm on the new UI) and currently it fails, I will try to adjust.

P.s. sorry for missing that some people still have the old UI, I didn't think about that

src/scrapers/base-beinleumi-group.ts
```diff
@@ -298,37 +304,47 @@ async function getAccountIdsBySelector(page: Page): Promise<string[]> {
...
-    accounts.push(accountData);
-  } else {
-    for (const accountId of accountsIds) {
+  
+  // Convert single account case to an array with one item
+  const accountsToProcess = accountsIds.length <= 1 ? [''] : accountsIds;
+  
+  for (const accountId of accountsToProcess) {
+    // Only select account if we have multiple accounts
+    if (accountsIds.length > 1) {
```
baruchiro, line 331: When you want the iteration to happened at least one time before testing the condition, you should use `do...while` loop.

baruchiro (commented):
@whatuserever @matanelgabsi @kapel2208 Please help us to test the new UI.

src/scrapers/base-beinleumi-group.ts
erikash: It's a condition to tell if you should select an account manually if there is more than one - 
it's parent statement a regular for each..
what am I missing?

erikash:
> @erikash please don't merge this yet, I test it (I'm on the new UI) and currently it fails, I will try to adjust. image
> 
> P.s. sorry for missing that some people still have the old UI, I didn't think about that

Thanks :)

src/scrapers/base-beinleumi-group.ts
nisan-tagar: @erikash maybe handle the conditional false case (no multi accounts) externally of the for loop and the for will only process the positive case:
1. If single account to process fetch account data return accounts.
2. Else, for on account Ids, fetch account data and return accounts.

You can add a small method for doing the getTransactionsFrame and getTransactionsFrame part serving both cases.

nisan-tagar:
> @erikash please don't merge this yet, I test it (I'm on the new UI) and currently it fails, I will try to adjust. image
> 
> P.s. sorry for missing that some people still have the old UI, I didn't think about that

Thanks!

src/scrapers/base-beinleumi-group.ts
erikash: It was like this before but I don't think the code is clearer.... the duplication you suggest in 2 feels worse than this if

erikash:
@nisan-tagar did you get a change to look into it?

src/scrapers/base-beinleumi-group.ts
baruchiro: Here is my proposal:

```typescript
async function selectAccount(page: Page, accountId: string) {
  await page.select('#account_num_select', accountId);
  await waitUntilElementFound(page, '#account_num_select', true);
}

async function fetchAccountDataBothUIs(page: Page, startDate: Moment) {
  // Try to get the iframe for the new UI
  const frame = await getTransactionsFrame(page);
    
  // Use the frame if available (new UI), otherwise use the page directly (old UI)
  const targetPage = frame || page;
  return fetchAccountData(targetPage, startDate);
}

async function fetchAccounts(page: Page, startDate: Moment):Promise<TransactionsAccount[]> {
  const accountsIds = await getAccountIdsBySelector(page);

  if (accountsIds.length <= 1) {
    const accountData = await fetchAccountDataBothUIs(page, startDate);
    return [accountData];
  }

  const accounts: TransactionsAccount[] = [];
  for (const accountId of accountsIds) {
    await selectAccount(page, accountId);
    
    const accountData = await fetchAccountDataBothUIs(page, startDate);
    accounts.push(accountData);
  }
  
  return accounts;
}
```

nisan-tagar: Thanks @baruchiro, the added method **fetchAccountDataBothUIs** resolves any duplications and I find the code more readable like this vs the conditional in the for loop.

nisan-tagar:
Thanks @erikash, 
I tested your branch and the scrapper seems to work now on my FIBI account UI that is currently using the old page format.

erikash:
We need someone on the new UI to test it

src/scrapers/base-beinleumi-group.ts
erikash: Ok, will update

erikash:
@Nic3Guy - did you get a change to make a fix? i'm assuming it would be really small
We need to merge this soon as the scraper is broken for users with the old UI

Nic3Guy:
@erikash I just tested it, worked on the new UI as well, thx

erikash:
@baruchiro added your suggested changes - please rereview
@Nic3Guy would be great if you can pull the changes and try again just to be sure :)

nisan-tagar (approved):

Issue #911, opened 2025-03-26
cal scraper is failing due to sms consent


sohodf:
hi, 
i'm using this project with moneyman scraper.
moneyman is now at 5.4.4, and scraping cal fails due to not being able to get past the consent to receiving messages pop-up(picture attached).
can this issue be fixed?
thanks ahead!

![Image](https://github.com/user-attachments/assets/4e4c3589-86a7-4437-9b8e-f705e40877fe)

daniel-hauser:
[5.4.4 includes the fix](https://github.com/eshaham/israeli-bank-scrapers/compare/v5.4.3...v5.4.4) and https://github.com/daniel-hauser/moneyman/pull/411 uses that version.

Can you share more on how you run moneyman?
If you have a fork, verify that you are fully synced. if you use the [moneyman docker container](https://github.com/daniel-hauser/moneyman/pkgs/container/moneyman) verify you have the latest version pulled

sohodf:
hi, no fork. nothing fancy.

after an update -
1. git pull
2. Run npm install
3. Run npm run build
4. Run npm run start

usually just 
npm run start

i just did that again to verify and got another consistent error - a timeout. unclear why.
might be the late hours. will try again tomorrow.

shaharzelig:
I am having the same problem. I am using Caspion.

shaiu:
I also had an issue in Caspion and I can confirm the PR that I published fixes it. I tried locally

maorsom:
I encountered the same issue — when I clicked "No thanks," the message disappeared, but the process still failed.

After looking into it, I found that the postAction hook tries to wait for the "Next" button, which never appears in this flow. As a result, it fails with the error: "wait for iframe with login form".

To handle this, the postAction hook should ideally support waiting for either the popup or allowing the process to continue without it. As a workaround, I wrapped waitUntilElementFound and waitForNavigation in a Promise.race, which solved the issue for me. However, I no longer see the popup in my flow, so I wasn’t able to fully test this across all scenarios.

Here’s the relevant part of the postAction hook:

```typescript
...
postAction: async () => {
  try {
    await Promise.race([
      waitForNavigation(this.page),
      waitUntilElementFound(this.page, 'button.btn-next-time'),
    ]);

    const currentUrl = await getCurrentUrl(this.page);
    ...

``` 

would be great to hear if anyone else can verify this workaround in different popup scenarios!

shayki5:
[...]

PR #913, opened 2025-03-28
fix(visa-cal): update button selector for next time button

shaiu:
Fix https://github.com/eshaham/israeli-bank-scrapers/issues/911

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -5,7 +5,7 @@ import {
   clickButton, elementPresentOnPage, pageEval, waitUntilElementFound,
 } from '../helpers/elements-interactions';
 import { fetchPostWithinPage } from '../helpers/fetch';
-import { getCurrentUrl } from '../helpers/navigation';
+import { getCurrentUrl, waitForNavigation } from '../helpers/navigation';
 import { getFromSessionStorage } from '../helpers/storage';
 import { filterOldTransactions } from '../helpers/transactions';
 import { waitUntil } from '../helpers/waiting';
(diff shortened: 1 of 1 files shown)


baruchiro:
For me, it is working with and without this change.

Anyone else?

@galbarm @daniel-hauser @idryzhov

maorsom:
Hi @baruchiro,

for me it doesn't work, I have a different scenario, in my case, the sms consent pop doesn't show up which cause the postAction to fail with "wait for iframe with login form"

I believe the core issue in my scenario is that the postAction assumes a consistent flow, but in some case( like mine) it needs to handle an alternate path.

I looked into this and it seems that the postAction hook should ideally support waiting for either the popup or allowing the process to continue without it. As a workaround, I wrapped waitUntilElementFound and waitForNavigation in a Promise.race, which solved the issue for me. However, I don't see the popup in my flow, so I wasn’t able to fully test this across all scenarios.

Here’s the relevant part of the postAction hook:

...
postAction: async () => {
  try {
    await Promise.race([
      waitForNavigation(this.page),
      waitUntilElementFound(this.page, 'button.btn-next-time'),
    ]);

    const currentUrl = await getCurrentUrl(this.page);
...

any thoughts?

baruchiro:
@maorsom Yes, this is good; we already use Promise.race for such cases.

But if you add the `waitForNavigation`, it will always return first because the popup on `waitUntilElementFound` happened after the page navigation. We need to think about how to know we passed the login, no matter if the pop-up is there or not.

Usually, we check the URL.

maorsom:
Thanks for the reply @baruchiro,

the question is, do we really need to wait for the popup next button? can't we just wait for the navigation to complete?

I see that the fetchData method is using an API callout rather then scraping the results from the page.

I don't see a reason to really close the popup, what I'm missing?

src/scrapers/visa-cal.ts
```diff
@@ -362,7 +362,7 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
       preAction: this.openLoginPopup,
       postAction: async () => {
         try {
-          await waitUntilElementFound(this.page, 'button.btn-next-time');
+          await waitUntilElementFound(this.page, 'button.butn-small-light-grey.btn-next-time');
```
baruchiro, line 365: I changed it to `waitForNavigation` and it worked for me, and I have the popup.

```suggestion
          await waitForNavigation(this.page);
```

baruchiro: I changed it, @maorsom @shaiu

PR #916, opened 2025-04-03
Isracard headless block fix: use `puppeteer-extra-plugin-stealth`

### LINKED ISSUE #915: Isracard blocks headless browser.
Scraping from isracard works only when showBrowser=True, otherwise it failed. When I took a screenshot I got the following:

Image

I tried curl isracard and got the same issue :(

Powershell example:
 $> Invoke-WebRequest -Uri digital.isracard.co.il/ -UseBasicParsing
output:
Invoke-WebRequest : The remote server returned an error: (403) Forbidden.
At line:1 char:12
+ $request = Invoke-WebRequest -Uri digital.isracard.co.il/ -UseBasicPa ...
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

gilgardosh:
close #915

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,8 @@
         "moment-timezone": "^0.5.37",
         "node-fetch": "^2.2.0",
         "puppeteer": "22.12.1",
+        "puppeteer-extra": "^3.3.6",
+        "puppeteer-extra-plugin-stealth": "^2.11.2",
         "uuid": "^9.0.1"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


baruchiro:
1. I want to be very careful with adding external packages, I will review these packages, and maybe I will take the user-agent and use only it.
2. You added it for all scrapers. I hope other banks are not blocking something that added by the "stealth".

Issue #933, opened 2025-05-05
max: support the new login step


daniel-hauser:
Max added a new step to the login screen that requires one more click for the first login in the browser

| Existing | New|
| -- | -- |
| ![Image](https://github.com/user-attachments/assets/660308b4-9e4b-475a-b2a8-b431580b02d2) | ![Image](https://github.com/user-attachments/assets/27a4f150-e74b-4f6d-a90a-ff33dc1d0cbb) |
|| ![Image](https://github.com/user-attachments/assets/2b223c68-df36-47fd-901d-e082ca950919) |

PR #934, opened 2025-05-05
fix(max): support the new max ui

daniel-hauser:
Fixes: #933

------
Tested in https://github.com/daniel-hauser/moneyman/pull/456


baruchiro (approved):
Worked for me too

PR #942, opened 2025-05-22
refactor: add prettier, remove babel

daniel-hauser:
This PR adds prettier and removes babel from the project.

## Prettier
- Prettier was added
- Overlapping eslint rules were removed
- Ran `npm run lint:fix` to format all files

## Babel
Since the project supports node >=18, babel is not needed, therefore safe to remove


package.json
```diff
@@ -21,7 +21,7 @@
     "build": "npm run lint && npm run clean && npm run build:js",
     "build:js": "tsc --project tsconfig.build.json",
     "postbuild": "rimraf lib/tests",
-    "prepare:core": "git reset --hard && node utils/prepare-israeli-bank-scrapers-core.js && npm i --package-lock-only && npm ci && npm run lint -- --fix && npm run build",
+    "prepare:core": "git reset --hard && node utils/prepare-israeli-bank-scrapers-core.js && npm i --package-lock-only && npm ci && npm run lint:fix && npm run build",
```
baruchiro, line 24: The code after "prepare:core" is not committed, I don't see a reason to lint:fix it.

daniel-hauser: That make sense, removing

PR #948, opened 2025-06-05
fix(otsar): Fixed otsar new UI

### LINKED ISSUE #921: Otsar Hahayal: still broken
Hello everyone!
FIrst of all, thanks for the wonderful project, I have been using it for above a year now.  :)

Unfortunately, since the Beinleumi bank changed their UI, the scraper seems to be broken for Otsar Hahayal (at least for me). I noticed the update at https://github.com/eshaham/israeli-bank-scrapers/pull/908, however it didn't seems to do the trick. The initial issue I could find is that the [success urls](https://github.com/eshaham/israeli-bank-scrapers/blob/7e10a58b0c9781f776e3caf9f651078922e3da10/src/scrapers/base-beinleumi-group.ts#L56) do not match the one in Otsar Hahayal (which doesn't have fibi at all).

Does anyone else experience this same issue? I tried modifying this locally but it seems like it is not the only fix needed, so I figured I should ask here.

kapel2208:
Fixes #921 

* Added the PortalNG login result page. This is the name in the new API.
* waitForNavigationAndDomLoad instead of waitForNavigation since accountId appears after DOM load.4

Tests are working locally on otsar account, probably need to manually test Binleumi as well?

### DIFF
--- a/src/helpers/navigation.ts
+++ b/src/helpers/navigation.ts
@@ -5,8 +5,8 @@ export async function waitForNavigation(pageOrFrame: Page | Frame, options?: Wai
   await pageOrFrame.waitForNavigation(options);
 }
 
-export async function waitForNavigationAndDomLoad(page: Page) {
-  await waitForNavigation(page, { waitUntil: 'domcontentloaded' });
+export async function waitForNavigationAndDomLoad(pageOrFrame: Page | Frame) {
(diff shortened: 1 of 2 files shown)


nisan-tagar:
@kapel2208 lets close this PR as we have the combined working changes already on https://github.com/eshaham/israeli-bank-scrapers/pull/949

PR #949, opened 2025-06-05
fix(beinleumi): add waits on balance and account elements

nisan-tagar:
Added wait statement on balance and account number element getters. 
After testing in some runtime environments (Github actions) these can render slower causing "fail to find element ... " errors that are not observed on local machine runtime.
Also cleaned up wait timeout in base scrapper and unified under a single const.

### DIFF
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -31,6 +31,7 @@ const PENDING_TRANSACTIONS_TABLE = 'table#dataTable023';
 const NEXT_PAGE_LINK = 'a#Npage.paging';
 const CURRENT_BALANCE = '.main_balance';
 const IFRAME_NAME = 'iframe-old-pages';
+const ELEMENT_RENDER_TIMEOUT_MS = 10000;
 
 type TransactionsColsTypes = Record<string, number>;
 type TransactionsTrTds = string[];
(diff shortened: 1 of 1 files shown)


nisan-tagar:
@kapel2208 this appears to be graceful your https://github.com/eshaham/israeli-bank-scrapers/pull/948 PR. 
@daniel-hauser I observed some errors ("Failed to find element matching selector ... ") while testing on Github Actions (moneyman) I think this should fix the issues.

kapel2208:
Thanks @nisan-tagar!
I indeed think I can remote the waitForNavigationAndDomLoad change I added.

After rebasing my commit to updated master (and this PR), the scraping works but I get a failure on browser.close() (closing the page will automatically clean up all frames). Have you seen such an error? I am still debuging it, but if you have a clue it can help.

nisan-tagar:
Hey @kapel2208, thanks for the feedback!
I checked out your [rebased branch](https://github.com/kapel2208/israeli-bank-scrapers/tree/kapel/fix-otsar-new-ui-rebased) (I see you revered the waitForNavigationAndDomLoad change).
I am able to build your branch and scrape my FIBI accounts, all tests pass and no failures observed. 
Since you can scrape OTSAR as well now with this branch combined changes perhaps we could pull it into this PR and close yours?

I see that browser.close() is happening in the base-scraper class (after the implementors actual scraping work) so perhaps this is a local infra issue you are facing? Maybe @baruchiro is familiar with this browser failure ...

daniel-hauser (approved):

nisan-tagar:
Thanks @daniel-hauser!
Updating that the FIBI scrapper (Beinleumi base scrapper) now works for me with these changes in 6.1.4 via Github Actions.

PR #955, opened 2025-06-30
fix: add page ready state check to prevent execution context errors

### LINKED ISSUE #952: Protocol error (Target.closeTarget): No target with given id found
Scraping returns:
```
{
  success: false,
  errorType: 'GENERIC',
  errorMessage: 'Protocol error (Target.closeTarget): No target with given id found'
}
```

Getting this since yesterday for all scrapers.
No code was changed. This is the code that was always working:

```
import { ScraperScrapingResult, createScraper } from 'israeli-bank-scrapers'

async function fetch(account: any, from: Date): Promise<ScraperScrapingResult> {
  const options = {
    companyId: account.company,
    startDate: from,
    combineInstallments: false,
    showBrowser: false,
    timeout: 120000,
    defaultTimeout: 120000
  }

  const credentials = {
    username: account.username,
    password: account.password,
    id: account.id,
    card6Digits: account.card6Digits,
    num: account.num,
    userCode: account.userCode
  }

  logger.info(`fetching company ${options.companyId}...`)

  const scraper = createScraper(options)
  const result = await scraper.scrape(credentials)

  logger.debug(result)  //No target with given id found'...
   ...
}
```

tomyweiss:
- Fix "Protocol error (Target.closeTarget): No target with given id found" caused by cleanup functions being called after page closure
- Fix "Navigation timeout of 30000 ms exceeded" caused by executing code before page is ready
- Add waitForFunction check to ensure document.readyState === 'complete' before $eval operations in pageEval function

Resolves #952, #962

### DIFF
--- a/src/helpers/elements-interactions.ts
+++ b/src/helpers/elements-interactions.ts
@@ -82,6 +82,7 @@ async function pageEvalAll<R>(
 ): Promise<R> {
   let result = defaultResult;
   try {
+    await page.waitForFunction(() => document.readyState === 'complete');
     result = await page.$$eval(selector, callback, ...args);
   } catch (e) {
     // TODO temporary workaround to puppeteer@1.5.0 which breaks $$eval bevahvior until they will release a new version.
@@ -102,6 +103,7 @@ async function pageEval<R>(
 ): Promise<R> {
   let result = defaultResult;
   try {
+    await pageOrFrame.waitForFunction(() => document.readyState === 'complete');
     result = await pageOrFrame.$eval(selector, callback, ...args);
   } catch (e) {
     // TODO temporary workaround to puppeteer@1.5.0 which breaks $$eval bevahvior until they will release a new version.
--- a/src/scrapers/base-scraper-with-browser.ts
+++ b/src/scrapers/base-scraper-with-browser.ts
@@ -75,6 +75,14 @@ function createGeneralError(): ScraperScrapingResult {
   };
 }
 
+async function safeCleanup(cleanup: () => Promise<void>) {
+  try {
+    await cleanup();
+  } catch (e) {
+    debug(`Cleanup function failed: ${(e as Error).message}`);
+  }
+}
+
 class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends BaseScraper<TCredentials> {
   private cleanups: Array<() => Promise<void>> = [];
 
@@ -291,7 +299,7 @@ class BaseScraperWithBrowser<TCredentials extends ScraperCredentials> extends Ba
       });
     }
 
-    await Promise.all(this.cleanups.reverse().map(cleanup => cleanup()));
+    await Promise.all(this.cleanups.reverse().map(safeCleanup));
     this.cleanups = [];
   }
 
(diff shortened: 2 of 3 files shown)


geriamir:
@eshaham sounds promising, can we merge this fix?

Amitkad:
@eshaham can this be merged? it blocks many people

src/scrapers/base-scraper.ts
```diff
@@ -56,7 +56,9 @@ export class BaseScraper<TCredentials extends ScraperCredentials> implements Scr
       const success = scrapeResult && scrapeResult.success === true;
       await this.terminate(success);
     } catch (e) {
-      scrapeResult = createGenericError((e as Error).message);
+      if (!(e as Error).message.includes('Target.closeTarget')) {
```
baruchiro: You're hiding the error here like you sure that error is not flagging we are doing something wrong.

Even if so, I want you to catch it and hide it inside the terminate function because it related to it.

src/helpers/elements-interactions.ts
```diff
@@ -82,6 +82,7 @@ async function pageEvalAll<R>(
 ): Promise<R> {
   let result = defaultResult;
   try {
+    await page.waitForFunction(() => document.readyState === 'complete');
```
baruchiro, line 85: Why? What if I want to catch it even if the page didn't loaded?

baruchiro:
@tomyweiss can you include my change in your?
#968 
It is somehow solved the problem of target closed without filtering that error in the catch block.

src/helpers/elements-interactions.ts
tomyweiss: I hadn’t thought about that—do we have any use cases where we’d want to catch things before the page finishes loading?
If so we can pass a variable that tells the eval request whether to wait for the page to fully load.

tomyweiss:
@baruchiro added your fix and removed the error throw.

src/helpers/elements-interactions.ts
baruchiro: You can check the usages, but you will never be sure because scrapers are fragile.

There are already enough variables to this function; it does what it must do.
If someone has a problem with the `readyState`, they can call that `page.waitForFunction(() => document.readyState === 'complete')` before (you may add it as a util too if you want to help the next person).

baruchiro: But I have a second thought. I don't know what triggered you to add this line, but if you think the absence of this line makes our scrapers more flaky, not in a specific case, but generally everywhere we use them, this can be something we need to consider.

I prefer adding this line to make our interactions more stable, above omitting it to make them faster.

src/scrapers/base-scraper-with-browser.ts
```diff
@@ -75,6 +75,14 @@ function createGeneralError(): ScraperScrapingResult {
   };
 }
 
+async function safeCleanup(cleanup: () => Promise<void>) {
+  try {
+    await cleanup();
+  } catch (e) {
+    debug(`Cleanup function failed: ${(e as Error).message}`);
+  }
```
baruchiro:
```suggestion
  } catch (e) {
    debug(`Cleanup function failed`, e);
  }
```

https://github.com/eshaham/israeli-bank-scrapers/pull/968#discussion_r2250087298

geriamir:
@tomyweiss can we fix the conflicts and merge this?

geriamir:
who can merge this now? :)

tomyweiss:
@baruchiro  seems like **6.2.0** did resolve the "No target with given id found" error, but I still see the "Navigation timeout" errors (specifically in the **Bank Otsar** scrapper).
Adding the `await page.waitForFunction(() => document.readyState === 'complete');` does solve it.

PR #964, opened 2025-07-28
fix(cal): authorization not being set

### LINKED ISSUE #963: Cal: Unexpected end of JSON input
Cal scraping is failing due to the following issue: Unexpected end of JSON input
I have found the issue.
The getAuthorizationHeader function is called before the token is set, causing a null value to be set as the token.
I have a fix, just want to confirm that others are experiencing this before opening a PR?

DavidConnack:
Changes:

* Fixed #963 

By fetching the auth header later in the process, it is populated and not null

### DIFF
--- a/src/helpers/elements-interactions.ts
+++ b/src/helpers/elements-interactions.ts
@@ -82,6 +82,7 @@ async function pageEvalAll<R>(
 ): Promise<R> {
   let result = defaultResult;
   try {
+    await page.waitForFunction(() => document.readyState === 'complete');
     result = await page.$$eval(selector, callback, ...args);
   } catch (e) {
     // TODO temporary workaround to puppeteer@1.5.0 which breaks $$eval bevahvior until they will release a new version.
@@ -102,6 +103,7 @@ async function pageEval<R>(
 ): Promise<R> {
   let result = defaultResult;
   try {
+    await pageOrFrame.waitForFunction(() => document.readyState === 'complete');
     result = await pageOrFrame.$eval(selector, callback, ...args);
   } catch (e) {
     // TODO temporary workaround to puppeteer@1.5.0 which breaks $$eval bevahvior until they will release a new version.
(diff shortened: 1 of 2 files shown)


DavidConnack:
I agree its not a perfect fix, but as with all things scraping, its a game of whackamole.
We could add a sleep, but in my testing this consistently fixed the issue (for now)

src/scrapers/visa-cal.ts
```diff
@@ -411,9 +411,9 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
     const startMoment = moment.max(defaultStartMoment, moment(startDate));
     debug(`fetch transactions starting ${startMoment.format()}`);
 
-    const Authorization = await this.getAuthorizationHeader();
     const cards = await this.getCards();
     const xSiteId = await this.getXSiteId();
+    const Authorization = await this.getAuthorizationHeader();
```
baruchiro: You can see how `this.getCards` using the `waitUntil` utility to wait until the value is there (or timeout).

You can use the same approach, and to save time, call both `getCards` and `getAuthorizationHeader` with `Promise.all`.

DavidConnack: @baruchiro thanks for the feedback done.
I found an issue with the base-scraper:
```
      const success = scrapeResult && scrapeResult.success === true;
      await this.terminate(success);
```
The terminate breaks the tests, but I don't think its related to this fix.
It seems to be related to this issues: #952

baruchiro:
I tested and it works, but I want to wait for #968

DavidConnack:
Thanks @baruchiro

ValYouW:
This doesn't seem to solve the issue in my case, maybe I didn't port the code correctly (I ported it directly to the transpiled javascript, not via typescript).

What happens is that `getFromSessionStorage` returns the following "truthy" value:
```js
{ auth: { calConnectToken: null, timestamp: null } }
```

This causes `waitUntil` to return although `authModule.auth.calConnectToken` is `null`.

DavidConnack:
@baruchiro 
Can you point me in the direction of them? I checked in the helpers, but I couldnt see anything relevant.
Thanks

PR #980, opened 2025-09-01
fix(visa-cal): better handling of authmodule waiting

DavidConnack:
The fix for  #963 was incomplete, see the discussion [here](https://github.com/eshaham/israeli-bank-scrapers/pull/964#issuecomment-3223260222).

This PR aims to fix the retry in `waitUntil` by checking for the actual token and not just a truthy value.

Included is a minor typo fix in the same file


kapel2208:
Validated locally and it is wokring for me

PR #991, opened 2025-10-05
fix(mizrahi): mark transactions as pending if they have no identifier

daniel-hauser:


src/scrapers/mizrahi.ts
```diff
@@ -260,6 +276,24 @@ class MizrahiScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
       balance: +response.body.fields?.Yitra,
     };
   }
+
+  private shouldMarkAsPending(txn: Transaction): boolean {
+    if (this.options.optInFeatures?.includes('mizrahi:pendingIfNoIdentifier') && !txn.identifier) {
```
baruchiro, line 380: make the `mizrahi:pendingIfNoIdentifier` a reusable constant

src/scrapers/mizrahi.ts
```diff
@@ -260,6 +276,24 @@ class MizrahiScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
...
+      debug(`Marking transaction '${txn.description}' as pending due to no identifier.`);
+      return true;
+    }
+
+    const genericDescriptions = ['העברת יומן לבנק זר מסניף זר'];
```
baruchiro: Put this in the head of the file, I think it is good to hold all the "scraping keys" in one place

src/scrapers/interface.ts
```diff
@@ -154,6 +159,11 @@ export type ScraperOptions = ScraperBrowserOptions & {
    * The number of times to retry the navigation in case of a failure (default 0)
    */
   navigationRetryCount?: number;
+
+  /**
+   * Opt-in features for the scrapers, allowing safe rollout of new breaking changes.
+   */
+  optInFeatures?: Array<OptInFeatures>;
```
baruchiro, line 166: It is a whole new feature... please document it in the readme.

src/scrapers/mizrahi.ts
daniel-hauser: I prefer letting TS verify the constants instead of defining consts (and enums have their own issues).
WDYT?

src/scrapers/interface.ts
daniel-hauser: Done

src/scrapers/mizrahi.ts
baruchiro: I think holding a string will require me to use the find tool, and it doesn't gurentee it will find only the related items, while using a reference, even a `const`, will send me by "go to declaration" to the list of the feature.

But I'm not really arguing on such things, this is my convention, but I accept yours too.

src/scrapers/interface.ts
```diff
@@ -22,6 +22,11 @@ export type ScraperCredentials =
         }
     ));
 
+export type OptInFeatures =
```
daniel-hauser, line 25: @baruchiro i like the idea of having `OptInFeatures`, but do you think my specific changes here are really opt-in and not bug fixes?

baruchiro: I see I already approved.

I didn't researched this. If you see it is a bug fix, remove the opt in feature.

Let me know please

PR #993, opened 2025-10-06
fix(isracard): add random delay for human-like login behavior

tomerroditi:
Implemented a helper function to introduce random delays during the login process, enhancing the scraping behavior to mimic human interaction, which resolves the automation block to login into isracard account.

### DIFF
--- a/src/scrapers/base-isracard-amex.ts
+++ b/src/scrapers/base-isracard-amex.ts
@@ -116,7 +116,12 @@ function getAccountsUrl(servicesUrl: string, monthMoment: Moment) {
   });
 }
 
-async function fetchAccounts(page: Page, servicesUrl: string, monthMoment: Moment, botFightingOptions?: BotFightingOptions): Promise<ScrapedAccount[]> {
+async function fetchAccounts(
+  page: Page,
+  servicesUrl: string,
+  monthMoment: Moment,
+  botFightingOptions?: BotFightingOptions,
+): Promise<ScrapedAccount[]> {
   const dataUrl = getAccountsUrl(servicesUrl, monthMoment);
   const dataResult = await fetchGetWithinPage<ScrapedAccountsWithinPageResponse>(page, dataUrl, botFightingOptions);
   if (dataResult && _.get(dataResult, 'Header.Status') === '1' && dataResult.DashboardMonthBean) {
@@ -331,7 +336,9 @@ function getExtraScrap(
   allMonths: moment.Moment[],
   botFightingOptions?: BotFightingOptions,
 ): Promise<ScrapedAccountsWithIndex[]> {
-  const actions = accountsWithIndex.map((a, i) => () => getExtraScrapAccount(page, options, a, allMonths[i], botFightingOptions));
+  const actions = accountsWithIndex.map(
+    (a, i) => () => getExtraScrapAccount(page, options, a, allMonths[i], botFightingOptions),
+  );
   return runSerial(actions);
 }
 
@@ -346,7 +353,8 @@ async function fetchAllTransactions(
   const allMonths = getAllMonthMoments(startMoment, futureMonthsToScrape);
   const results: ScrapedAccountsWithIndex[] = await runSerial(
     allMonths.map(
-      monthMoment => () => fetchTransactions(page, options, companyServiceOptions, startMoment, monthMoment, botFightingOptions),
+      monthMoment => () =>
+        fetchTransactions(page, options, companyServiceOptions, startMoment, monthMoment, botFightingOptions),
     ),
   );
 
@@ -453,7 +461,13 @@ class IsracardAmexBaseScraper extends BaseScraperWithBrowser<ScraperSpecificCred
         countryCode: COUNTRY_CODE,
         idType: ID_TYPE,
       };
-      const loginResult = await fetchPostWithinPage<{ status: string }>(this.page, loginUrl, request, {}, this.options.botFightingOptions);
+      const loginResult = await fetchPostWithinPage<{ status: string }>(
+        this.page,
+        loginUrl,
+        request,
+        {},
+        this.options.botFightingOptions,
+      );
       debug(`user login with status '${loginResult?.status}'`);
 
       if (loginResult && loginResult.status === '1') {
--- a/src/helpers/anti-automation-detection.ts
+++ b/src/helpers/anti-automation-detection.ts
@@ -11,10 +11,7 @@ export function randomMouseMove(page: Page): Promise<void> {
   return page.mouse.move(x, y);
 }
 
-export default async function fightBotDetection(
-  page: Page,
-  options: BotFightingOptions,
-): Promise<void> {
+export default async function fightBotDetection(page: Page, options: BotFightingOptions): Promise<void> {
   const withRandomDelay = options?.withRandomDelay ?? false;
   const withMouseMove = options?.withMouseMove ?? false;
 


src/scrapers/base-isracard-amex.ts
```diff
@@ -391,21 +345,28 @@ class IsracardAmexBaseScraper extends BaseScraperWithBrowser<ScraperSpecificCred
   }
 
   async login(credentials: ScraperSpecificCredentials): Promise<ScraperScrapingResult> {
+    // Helper for random human-like delays
+    const randomDelay = async (min = 500, max = 1500) => {
```
daniel-hauser: Can `randomDelay` be moved to `src/helpers/waiting.ts`? also, the existing `sleep` can be used for the delayed promise

tomerroditi: sure, i minimized the MR to use `sleep`, tested with a single call instead of 2 calls and it works fine as well.
btw i noticed that the error of blocking automation happens only after 3-4 repeated attempts of scraping (without the added sleep delay)

liraneli: the same should also be done in the fetchData function as it also do simultaneous calls to the API

src/scrapers/base-isracard-amex.ts
```diff
@@ -392,19 +346,19 @@ class IsracardAmexBaseScraper extends BaseScraperWithBrowser<ScraperSpecificCred
...
       } else {
-        void request.continue(undefined, interceptionPriorities.continue);
+        void request.continue();
       }
     });
 
-    await maskHeadlessUserAgent(this.page);
-
+    debug('navigate to login page');
     await this.navigateTo(`${this.baseUrl}/personalarea/Login`);
```
daniel-hauser, line 421:
```suggestion
    await this.navigateTo(`${this.baseUrl}/personalarea/Login`, 'networkidle0');
```

Can you see if this change works instead of calling `sleep`? (`'networkidle2'` might also work, but `'networkidle0'` will wait more time)

liraneli: i don't think this change is related to network idle issue... it's due to blocks on isracard side

tomerroditi:
```
    src/scrapers/base-isracard-amex.ts:362:65 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'Page'.

    362     await this.navigateTo(`${this.baseUrl}/personalarea/Login`, 'networkidle0');
```

daniel-hauser: > await this.navigateTo(`${this.baseUrl}/personalarea/Login`,

interesting why, the signature of `navigateTo` has no `Page` in it

```ts
  async navigateTo(
    url: string,
    waitUntil: PuppeteerLifeCycleEvent | undefined = 'load',
    retries = this.options.navigationRetryCount ?? 0,
  ): Promise<void>
```

src/helpers/fetch.ts
```diff
@@ -52,59 +52,77 @@ export async function fetchGraphql<TResult>(
...
-              // No content response
-              resolve(null);
-            } else {
-              resolve(result.json());
-            }
-          })
-          .catch(e => {
-            reject(e);
-          });
+export async function safeFetchGetWithinPage<T>(page: Page, url: string): Promise<T | null> {
```
daniel-hauser: Considre adding a third optional `withRandomDelay = false` param to `fetchGetWithinPage` instead of adding a new `safeFetchGetWithinPage`. (yes, we will lose the nice `Automation blocked` error, but i think this error shouled not be handled in a generic fetch function)

[and same for `safeFetchPostWithinPage`]

src/scrapers/base-isracard-amex.ts
tomerroditi: maybe we are on other versions (i'm working with the latest version)?
```
async navigateTo(
    url: string,
    page?: Page,
    timeout?: number,
    waitUntil: PuppeteerLifeCycleEvent | undefined = 'load',
  ): Promise<void>
  ```

Comment (daniel-hauser) [reply src/scrapers/base-isracard-amex.ts:421]: > maybe we are on other versions (i'm working with the latest version)?

I removed `Page` in April as part of #928

Comment (tomerroditi) [reply src/scrapers/base-isracard-amex.ts:421]: my fork wasn't synced, retried after updating it, it doesn't solves the issue. i tiried both "networkidle2" and "networkidle0"

File: src/helpers/fetch.ts (hunk 3 above)
Comment (tomerroditi) [reply src/helpers/fetch.ts]: added, also added `withMouseMove` flag to enable random mouse movements.

**Comment by tomerroditi** [comment]
1. i'm sorry for the mess, first time writing type/javascript
2. synced my fork and rebased my work on top of it (was 1 year behind...)
3. implemented above suggestions to improve code structure and minimize additional code
4. added a new helpers module called "anti-automation-detection", feel free to suggest another name or file to place the utils functions in there
5. added random mouse moves, @daniel-hauser would love to hear if it solves the issue for you as well, mine works without it, I have 3 cards in my account, and I'm scraping 12 months back.

File: src/helpers/fetch.ts [hunk 4]
```diff
@@ -51,48 +53,30 @@ export async function fetchGraphql<TResult>(
...
+  if (withRandomDelay) {
+    promise = promise.then(() => randomDelay()).then(() => void 0);
+  }
+
+  if (withMouseMove) {
+    promise = promise.then(() => randomMouseMove(page));
+  }
+
+  return promise.then(() =>
+    page.evaluate(innerUrl => {
```
daniel-hauser: Changing the function to be `async` will allow you to simplify the code to:

```suggestion
  if (withRandomDelay) {
    await randomDelay();
  }

  if (withMouseMove) {
    await randomMouseMove(page);
  }

  return page.evaluate(innerUrl => {
```

src/helpers/fetch.ts
```diff
@@ -51,48 +53,30 @@ export async function fetchGraphql<TResult>(
...
-  }, url);
-}
-
-export function fetchPostWithinPage<TResult>(
+export function fetchGetWithinPage<TResult>(
   page: Page,
   url: string,
-  data: Record<string, any>,
-  extraHeaders: Record<string, any> = {},
+  withRandomDelay: boolean = false,
```
daniel-hauser: Considre changing to a `options` object instead of boolean args, this will make the callers code more readable (since JS dosn't have named parameters yet)

```ts
export async function fetchGetWithinPage<TResult>(
  page: Page,
  url: string,
  botFightingOptions?: BotFightingOptions,
): Promise<TResult | null> {
```

Then, you can expose the type and a single function from 

```ts
export type BotFightingOptions = {
  withRandomDelay?: boolean;
  withMouseMove?: boolean;
};

export async function fightBotDetection(
  page: Page,
  { withRandomDelay = false, withMouseMove = false }: BotFightingOptions = {},
): Promise<void> {
  if (withRandomDelay) {
    await randomMouseMove(page);
  }
  if (withMouseMove) {
    await randomDelay();
  }
}
```

and the call here can simply be 

```ts
await fightBotDetection(botFightingOptions);
```

daniel-hauser:
> 1. i'm sorry for the mess, first time writing type/javascript

Excellent job! the code doesn't show it 👌

> 5. added random mouse moves, @daniel-hauser would love to hear if it solves the issue for you as well, mine works without it, I have 3 cards in my account, and I'm scraping 12 months back.

The failures are sporadic, currently it works with the code in master. once i see failures i can test to see if it solves them

> 6. an important part of preventing the automation detection is to run each month data scraping sequentially rather that in parallel ([72b454a](https://github.com/eshaham/israeli-bank-scrapers/commit/72b454a484261004d32973ba05b1b40f78106695)) - we still need the additional delays to prevent detection but without serial run the delays are not helpful.
> 7. the delays and sequential run makes it very slow to scrape isracard, we might be able to smartly decide when we need to use them according to the amount of months and accounts we are scraping.

In #991 i proposed a new `optInFeatures` option that could have been useful here as well, but from the Moneyman POV (daily scrapes of the last 10 days, one/two `monthMoment`) I'm fine with the sequence

> from my tests, scraping up to 6 months works well without changing anything in the code. I would like to verify with other users if they are experiencing the same thing before starting to implement it.

It failed for me also when running with startDate of 10 days ago and futureMonthsToScrape=true

src/helpers/fetch.ts
tomerroditi: awesome, I added the anti detection options as an interface (i thought it might be easier to understant its options this way, let me know if that's fine), and made it possible to pass it as one of the fields in the scraper options. currently only used within base-isracard-amex scraper.

src/helpers/fetch.ts
tomerroditi: simplified it with async, and moved all anti detection logic to the helper module for better code modularity and SOC.

src/scrapers/base-isracard-amex.ts
```diff
@@ -334,17 +340,18 @@ async function fetchAllTransactions(
...
   const futureMonthsToScrape = options.futureMonthsToScrape ?? 1;
   const allMonths = getAllMonthMoments(startMoment, futureMonthsToScrape);
-  const results: ScrapedAccountsWithIndex[] = await Promise.all(
-    allMonths.map(async monthMoment => {
-      return fetchTransactions(page, options, companyServiceOptions, startMoment, monthMoment);
-    }),
+  const results: ScrapedAccountsWithIndex[] = await runSerial(
+    allMonths.map(
+      monthMoment => () => fetchTransactions(page, options, companyServiceOptions, startMoment, monthMoment, botFightingOptions),
+    ),
```
tomerroditi, line 358: should I insert an option to control this behavior via the anti bot detection options?

when we have many months it might slow down things but in that case it is almost certain to be caught as a bot, while having only a few months shouldn't suffer much from not running in parallel.

PR #1005, opened 2025-11-19
refactor(max): update LOGIN_URL to point to the login page

idofl:
Current URL (/homepage/welcome) fails with "GENERIC" error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,13 @@ const scrapeResult = await scraper.scrape({ username: 'vr29485', password: 'some
 await browser.close();
 ```
 
+### OptIn Features
+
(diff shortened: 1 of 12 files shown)


daniel-hauser:
I do not get any error when scraping max, can you share more on the error?
if you choose to update the URL of the login page, consider changing also `preAction` and `checkReadiness` that are currently used to click on the button that leads to the login page and are probably obsolete when opening the login page directly.

erikash:
@daniel-hauser @idofl I can confirm that max was broken for me and your PR @idofl fixes it

Please address @daniel-hauser 's comments and let's merge it

PR #1012, opened 2025-11-29
fix(discount): scrape transactions for all user accounts

### LINKED ISSUE #1011: Support multi account scraping for Bank Discount
Currently the scraper returns only the default account

Arielgordon123:
closes #1011

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "1.0.4",
       "license": "MIT",
       "dependencies": {
-        "build-url": "^2.0.0",
         "core-js": "^3.1.4",
         "debug": "^4.3.2",
         "lodash": "^4.17.10",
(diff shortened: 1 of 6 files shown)


baruchiro:
Hi, next time tag me or send me a whatsapp

src/scrapers/discount.ts
```diff
@@ -73,42 +78,47 @@ async function fetchAccountData(page: Page, options: ScraperOptions): Promise<Sc
       errorMessage: 'failed to get account data',
     };
   }
-  const accountNumber = accountInfo.UserAccountsData.DefaultAccountNumber;
 
   const defaultStartMoment = moment().subtract(1, 'years').add(2, 'day');
```
baruchiro, line 82: Interesting, why's the previous developer used the `DefaultAccountNumber`, if the response includes all acounts?

I'm trying to say, we need someone with only one account to test your change, maybe for one account only, the `UserAccounts` is not exists or empty?

yyogo: Hi, just tested, works with one account

PR #1020, opened 2025-12-14
feat(visa-cal): integrate frame (misgeret) data into fetchData

tomerh2001:
Added a fetch for frame (misgeret) data and incorporated it into the final result of `fetchData`.

The final result now has one new fields:
* `balance` = `-<current misgeret value>`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@tomerh2001/israeli-bank-scrapers",
+  "name": "israeli-bank-scrapers",
   "version": "1.0.4",
   "private": true,
   "description": "Provide scrapers for all major Israeli banks and credit card companies",
@@ -28,7 +28,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/tomerh2001/israeli-bank-scrapers.git"
+    "url": "git+https://github.com/eshaham/israeli-bank-scrapers.git"
   },
   "keywords": [
     "israel",
@@ -38,9 +38,9 @@
   "author": "Elad Shaham",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/tomerh2001/israeli-bank-scrapers/issues"
+    "url": "https://github.com/eshaham/israeli-bank-scrapers/issues"
   },
-  "homepage": "https://github.com/tomerh2001/israeli-bank-scrapers#readme",
+  "homepage": "https://github.com/eshaham/israeli-bank-scrapers#readme",
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,64 +2,64 @@ name: Release
 on:
   push:
     branches:
-    - master
-    - next
+      - master
+      - next
 jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout
-      uses: actions/checkout@v4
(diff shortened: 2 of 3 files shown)


src/scrapers/visa-cal.ts
```diff
@@ -502,6 +519,8 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
 
         return {
           txns,
+          frame,
```
Skeli789: Did you test this? Because you also need to add `frame` to `TransactionsAccount` in `src/transactions.ts` or this won't compile.

tomerh2001: Ngl it was just a nice bonus, I already set the correct value in `accountBalance`, so I will remove `frame`.

To your question, yes i did test it, I am actually using it in my home server.

tomerh2001: Done

package.json
```diff
@@ -78,10 +78,10 @@
     "core-js": "^3.1.4",
     "debug": "^4.3.2",
     "lodash": "^4.17.10",
-    "moment": "^2.22.2",
+    "moment": "^2.30.1",
     "moment-timezone": "^0.5.37",
     "node-fetch": "^2.2.0",
-    "puppeteer": "22.15.0",
+    "puppeteer": "^22.15.0",
```
baruchiro: why did you touched the puppeteer package?

package.json
```diff
@@ -78,10 +78,10 @@
     "core-js": "^3.1.4",
     "debug": "^4.3.2",
     "lodash": "^4.17.10",
-    "moment": "^2.22.2",
+    "moment": "^2.30.1",
```
baruchiro: I can't see where you handle date/time that will require you to upgrade the package.

src/scrapers/visa-cal.ts
```diff
@@ -502,6 +519,7 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
 
         return {
           txns,
+          balance: -frame?.nextTotalDebit,
```
baruchiro, line 519: If `frame` is `underfined`, the `balance` will be `NaN`. Please be more specific like:


```suggestion
          balance: _.isNumber(frame?.nextTotalDebit) ? -frame?.nextTotalDebit : undefined,
```

src/scrapers/visa-cal.ts
```diff
@@ -437,12 +440,26 @@ class VisaCalScraper extends BaseScraperWithBrowser<ScraperSpecificCredentials>
...
+    debug(`fetch frames (misgarot) of cards`);
+    const frames = await fetchPost(
+      FRAMES_REQUEST_ENDPOINT,
+      { cardsForFrameData: cards.map(({ cardUniqueId }) => ({ cardUniqueId })) },
+      {
+        Authorization,
+        'X-Site-Id': xSiteId,
+        'Content-Type': 'application/json',
+        ...apiHeaders,
+        Origin: 'https://digital-web.cal-online.co.il',
```
daniel-hauser: Can you replace the existing `Origin` and `Referer` on `apiHeaders` with the 'https://digital-web.cal-online.co.il' domain? i checked and now the site uses it in all api calls

tomerh2001: Done

package.json
tomerh2001: Done

package.json
tomerh2001: Done

baruchiro (approved):

Issue #1021, opened 2025-12-16
Hapoalim scraper returns "failed to fetch"


roidital:
Hi, I encountered this error since yesterday:
"success": false
"errorType": "GENERIC"
"errorMessage": "Failed to fetch"

when trying to scrape Hapoalim account.

manual login (from browser) - works fine.

DavidConnack:
I have the same, I have opened a PR with a fix for this

PR #1022, opened 2025-12-17
fix(hapoalim): only scrape account if it is an active account

DavidConnack:
In a situation where you have 2 accounts, one active and one inactive the current logic will still try and get the inactive accounts transactions causing the scrape to fail.
This change will call `getAccountTransactions` on active accounts

Resolves issue #1021

### DIFF
--- a/src/scrapers/hapoalim.ts
+++ b/src/scrapers/hapoalim.ts
@@ -197,7 +197,11 @@ async function fetchAccountData(page: Page, baseUrl: string, options: ScraperOpt
   debug('fetching accounts data');
   const accountsInfo = (await fetchGetWithinPage<FetchedAccountData>(page, accountDataUrl)) || [];
   const openAccountsInfo = accountsInfo.filter(account => account.accountClosingReasonCode === 0);
-  debug('got %d open accounts from %d total accounts, fetching txns and balance', openAccountsInfo.length, accountsInfo.length);
+  debug(
+    'got %d open accounts from %d total accounts, fetching txns and balance',
+    openAccountsInfo.length,
+    accountsInfo.length,
+  );
 
   const defaultStartMoment = moment().subtract(1, 'years').add(1, 'day');
   const startDate = options.startDate || defaultStartMoment.toDate();


src/scrapers/hapoalim.ts
```diff
@@ -209,31 +209,31 @@ async function fetchAccountData(page: Page, baseUrl: string, options: ScraperOpt
...
 
     const isActiveAccount = account.accountClosingReasonCode === 0;
     if (isActiveAccount) {
       balance = await getAccountBalance(apiSiteUrl, page, accountNumber);
+      const txns = await getAccountTransactions(
+        baseUrl,
+        apiSiteUrl,
+        page,
+        accountNumber,
+        startDateStr,
```
baruchiro: So now we have nothing to do with inactive accounts at all, do you think we even need to return them as empty?

Or maybe, we should find the way to support them? Thinking about that, if it was warking before (is it?) and now it failed, it is a breaking change.

No matter what we decide, I think the code right now should be simpler, since we don't need the inactive account, we should filter it from the loop (the ugliest way will be `if (!isActiveAccount) continue;`, but rearrange the code a little bit to find a better way.

daniel-hauser:
> @daniel-hauser I want to read your opinion

I do agree that it's a breaking change, but hapoalim are the ones that broke it. IMHO users that closed an account will understand why it is not scraped anymore so i'm fine with it

src/scrapers/hapoalim.ts
daniel-hauser: I do agree that with this change an early return (or even a `.filter`) will result in nicer code

baruchiro:
@erikash we are waiting for @DavidConnack to answer the review, you're welcome to share your opinion too (or edit the code according to the review)

DavidConnack:
> @erikash we are waiting for @DavidConnack to answer the review, you're welcome to share your opinion too (or edit the code according to the review)

Sorry missed this.
Will answer above

src/scrapers/hapoalim.ts
DavidConnack: Will take a look this week.
Thanks for the feedback

PR #1025, opened 2025-12-19
fix(discount): support new redirect  url

dankonsdev:
discount added a new redirect url, i just added it to the list to make the scraper work.


dankonsdev:
@Arielgordon123 can we merge it? it will really help my scraping flow from discount

src/scrapers/discount.ts
```diff
@@ -137,6 +137,7 @@ function getPossibleLoginResults(): PossibleLoginResults {
   urls[LoginResults.Success] = [
     `${BASE_URL}/apollo/retail/#/MY_ACCOUNT_HOMEPAGE`,
     `${BASE_URL}/apollo/retail2/#/MY_ACCOUNT_HOMEPAGE`,
+    `${BASE_URL}/apollo/retail2/`,
```
baruchiro, line 140: I wonder if the new one includes the previous one

PR #1031, opened 2026-01-04
feat: add option to include raw transaction data

daniel-hauser:
Introduce an option in the scraper settings to include raw transaction data for debugging.


daniel-hauser:
@baruchiro can you review? this is a opt-in feature with no effect unless `includeRawTransaction: true` is passed

PR #1068, opened 2026-03-11
fix(visa-cal): fix change password detection

gczobel:
## Problem

When a visaCal user's password needs to be changed, Cal's website shows an error popup after login prompting the user to reset their password via the login screen. The scraper was returning `TIMEOUT` instead of `CHANGE_PASSWORD`.

## Root cause

`hasChangePasswordForm()` was checking for `.change-password-subtitle`, which no longer exists on Cal's login page. The password change prompt now renders as an `.err-desc` element inside an `<error>` component:

```html
<error class="ng-star-inserted">
  <div class="err-wrapper">
    <div class="err-title">...</div>
    <div class="err-desc">... change password message ...</div>
  </div>
</error>
```

## Fix

- Replace the `.change-password-subtitle` selector with `.err-desc`
- Verify the element text includes the change-password phrase to avoid false positives from other error types that use the same element

## Tested

Verified with a real visaCal account that requires a password change. Before the fix: `TIMEOUT`. After the fix: `CHANGE_PASSWORD`.


gczobel:
Any codeowner can check this?

@baruchiro @galbarm @daniel-hauser @idryzhov @geriamir

baruchiro (approved):

PR #1070, opened 2026-03-12
fix(beinleumi): handle PortalNG post-login UI in waitForPostLogin

### LINKED ISSUE #1069: Otsar HaHayal: bank migrated to www.bankotsar.co.il, old online.bankotsar.co.il no longer works
## Description
The OtsarHaHayal scraper fails after successful login with:
```
Waiting for selector `#card-header` failed: Waiting failed: 30000ms exceeded
```

## Root cause
The bank has migrated from `online.bankotsar.co.il` to `www.bankotsar.co.il/private/`.

The current scraper uses:
- `BASE_URL = 'https://online.bankotsar.co.il'`
- `LOGIN_URL = '...online.bankotsar.co.il/MatafLoginService/...'`

When navigating to the old URL, login credentials are accepted but the session is immediately invalidated (redirects back to login), so `waitForPostLogin` never finds `#card-header` or any other post-login selector.

## Steps to reproduce
1. Configure `CompanyTypes.otsarHahayal` with valid credentials
2. Run scraper with `showBrowser: true`
3. Observe: login form submits successfully, then immediately redirects back to login page
4. Scraper times out waiting for `#card-header`

## Expected behavior
The URLs in `otsar-hahayal.ts` should point to the new portal at `www.bankotsar.co.il`.

## Environment
- `israeli-bank-scrapers` v6.7.1
- Node.js 22
- macOS (also tested on ubuntu-latest via GitHub Actions)

urihersh:
## Problem

Closes #1069

The bank migrated to a new PortalNG UI which no longer has the DOM selectors (`#card-header`, `#account_num`, etc.) that `waitForPostLogin` was waiting for. After a successful login the page navigates to a PortalNG URL, but none of the four `waitUntilElementFound` calls ever resolve — they all time out and the first rejection causes the scraper to fail with:

```
Waiting for selector `#card-header` failed
```

## Fix

**`waitForPostLogin`** — Add a URL-based detector as the primary check. It polls `page.url()` every 300 ms and resolves as soon as the page navigates to a known post-login URL pattern (`PortalNG/shell`, `FibiMenu/Online`, `accountSummary`). The existing DOM element checks are kept as fast-path fallbacks for the old UI, wrapped with `neverReject` so a timeout on any of them cannot pre-empt the URL detector.

**`fetchAccountData`** — Wrap in a try/catch so that if the new PortalNG UI doesn't expose the old transaction/balance selectors, the scraper returns empty data gracefully instead of crashing.

## Testing

Verified end-to-end against a live OtsarHahayal account — login, balance, investments, mortgage, credit card, and 22 transactions all returned correctly.

### DIFF
--- a/src/scrapers/base-beinleumi-group.ts
+++ b/src/scrapers/base-beinleumi-group.ts
@@ -312,25 +312,67 @@ async function getCurrentBalance(page: Page | Frame): Promise<number> {
 }
 
 export async function waitForPostLogin(page: Page) {
+  // URL-based detector: resolves when the page navigates to a known post-login URL.
+  // Handles the PortalNG new UI which doesn't have the old DOM selectors.
+  const urlDetected = new Promise<void>((resolve, reject) => {
+    const deadline = Date.now() + 300000;
+    const check = () => {
+      try {
+        const url = page.url();
+        if (/fibi.*accountSummary|Resources\/PortalNG\/shell|FibiMenu\/Online/.test(url)) {
+          resolve();
+          return;
+        }
+      } catch {
+        // page may not be ready yet
+      }
+      if (Date.now() >= deadline) {
+        reject(new Error('waitForPostLogin: post-login URL not detected within timeout'));
+        return;
+      }
+      setTimeout(check, 300);
+    };
+    check();
+  });
+
+  // Element-based checks are kept as fast-path fallbacks for the old UI.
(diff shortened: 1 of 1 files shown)


blorby:
pagaz, can i help some how to get this merged?

urihersh:
> pagaz, can i help some how to get this merged?

I don't know how it works in this repo. @eshaham can you please guide me if there is anything I should do?

Issue #1055, opened 2026-02-22
isracard scrap fail


moshe212:
<img width="491" height="192" alt="Image" src="https://github.com/user-attachments/assets/943b02d7-6682-4c1d-991e-eaf3bf14fe81" />

he not go to login with persist password

Issue #1053, opened 2026-02-02
Isracard scraper fails


Liron42:
The Isracard scraping fails with:
`Unexpected token 'B', "Block Automation" is not valid JSON`

I saw https://github.com/eshaham/israeli-bank-scrapers/issues/1045 - tried to clean puppetteer caches but the issue persists.. any ideas?

ramikachlon:
The real issue here is that Isracard's website detected the scraper as a bot and returned "Block Automation" instead of JSON. This is Isracard's anti-automation detection actively blocking the request

D3SL:
What's weird is I can usually get ONE isracard account to work but not the second.

D3SL:
This is not stale, bad bot.

Amitkad:
it is not stale. still happens

dadarara:
Workaround that fixed this for me, with no code change: **point `executablePath` at a real installed browser instead of the Chromium that gets downloaded automatically.**

## Symptom

Failure happened at the very first navigation, before any credential was sent:

```
GENERIC: Failed to navigate to url https://digital.isracard.co.il/personalarea/Login, status code: 403
```

~9 seconds, then the importer failed. Not a credentials problem — login was never attempted.

## What fixed it

Setting the browser executable to the Microsoft Edge already installed on the machine:

```
C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe
```

Isracard then logged in and scraped normally, 76 transactions on the first successful run. Nothing else changed — same machine, same credentials, same version of the scrapers library.

## Why this seems to work

The auto-downloaded Chromium reports client hints with no real-browser brand:

```
sec-ch-ua: "Chromium";v="…", "Not.A/Brand";v="…"
```

An installed Chrome or Edge includes its own brand alongside those (`"Google Chrome"` / `"Microsoft Edge"`), which appears to be enough to get past the check at the login page. Edge works because it is Chromium underneath, so puppeteer drives it unchanged.

## Things worth knowing

- **Running headed did not help on its own.** I ran with the browser visible throughout and still got the 403. Browser *identity* was the blocker, not `navigator.webdriver` — which means for this particular 403 the stealth flags may not be needed at all.
- This is the **login-page 403**, which is not quite the same failure as the `"Block Automation"` / 429 responses reported here and in #1098 — those get past login and are blocked on API calls afterwards. This workaround may or may not help those; I could not test it, as I never reached that stage.
[...]

Issue #997, opened 2025-10-13
Isracard scraper fails


liranay:
I'm getting the following error:
isracard scrape failed. errorType: GENERIC. errorMessage: Failed to execute json on Response: Unexpected end of JSON input
when showing the UI while scraping, the credentials are not entered and the browser exists

<img width="1024" height="1230" alt="Image" src="https://github.com/user-attachments/assets/2e76f57e-b305-464a-a97b-01bd18b324f1" />

Did anyone experience that issue?
Thanks

daniel-hauser:
I suspect that the issue is duplicate of #970 and not an issue with the login UI. the login UI is not used in `base-isracard-amex.ts`, we use a POST request to login.
I didn't get any error lately when scraping isracrad/amex cards, and the only issues where automation blocking issues.

benjhanbs:
> > i get this issue as well - did anyone solved it?
> 
> My fix in [#1027](https://github.com/eshaham/israeli-bank-scrapers/pull/1027) works for me, you just need to make sure you're not scraping additonal transaction information because then the scraper will get bot-blocked sending individual requests for each transaction.

I tried to scrape only Isracard, and it also didn't work.
Can someone push your code so we can all enjoy your fix?
How can I implement it in my code?

PR #1064, opened 2026-03-04
fix: bypass Cloudflare WAF blocking on Amex/Isracard login

eyaleng:
The Amex (and Isracard) scraper login fails with a Cloudflare "Sorry, you have been blocked" HTML page instead of the expected JSON response. This happens because Cloudflare WAF fingerprints headless Chrome via missing sec-ch-ua client hints and navigator.webdriver detection.

Replace the insufficient `maskHeadlessUserAgent` (which only stripped "HeadlessChrome" from the UA string) with:

1. A realistic User-Agent derived from the actual Chrome version
2. sec-ch-ua / sec-ch-ua-mobile / sec-ch-ua-platform HTTP headers injected via request interception on every outgoing request
3. JavaScript stealth overrides (navigator.webdriver, plugins, languages, chrome.runtime) applied before any page script runs

Fixes #1057

PR #1082, opened 2026-03-26
fix(isracard): rewrite scraper to bypass WAF using XLSX download 

OriStav:
Problem                                                                                                              
                                                                                                                       
  The Isracard scraper fails because Akamai WAF blocks all JSON API requests (/services/ProxyRequestHandler) with
  "Block Automation" and HTTP 429. The existing detector-dom.min.js blocking and header injection are no longer        
  sufficient to bypass the WAF. 
                                                                                                                       
  Related issues: #997, #1053, #1055
                                      
  Solution

  Rewrites the Isracard scraper (isracard.ts) to use a completely different data retrieval approach while keeping the  
  base class (base-isracard-amex.ts) untouched for Amex compatibility:
                                                                                                                       
  - Browser-based login — fills the password form via Puppeteer instead of calling the ValidateIdData/performLogonI API
   endpoints                          
  - XLSX download — navigates to the transactions page and downloads the Excel export instead of calling the blocked   
  JSON API                      
  - navigator.webdriver masking — hides the automation flag that WAF fingerprinting checks for
  - Multi-card support — new cardSuffixes option (last 4 digits per card) to scrape multiple cards; if omitted, scrapes
   the default card only
                                                                                                                       
  Files changed                 
                                      
  - src/scrapers/isracard.ts — Rewritten with XLSX-based scraping                                                      
  - src/scrapers/interface.ts — Added optional cardSuffixes option
  - package.json — Added xlsx dependency                                                                               
  - test-isracard.js — Manual test script
                                                                                                                       
  What's NOT changed
                                                                                                                       
  - base-isracard-amex.ts — untouched, Amex scraper continues to work as before
  - Transaction output format — returns standard Transaction[] objects
  - Scraper interface — no breaking changes
                                                                                                                       
  Context
                                                                                                                       
  Tested and validated against Isracard's current WAF configuration (March 2026). The approach was first developed in
  Clarify, which uses this library.   

  For maximum stealth, users can also pass a puppeteer-extra browser with StealthPlugin via options.browser. See also  
  @sergienko4's fork which uses Camoufox for C++-level anti-detection.
                                                                                                                       
  Test plan                     
                                      
  - Isracard login succeeds without WAF block                                                                          
  - Transactions are fetched and parsed correctly from XLSX
  - Multi-card scraping works with cardSuffixes option                                                                 
  - Single default card works when cardSuffixes is omitted
  - Amex scraper still works (shares base class, no changes)
  - No breaking changes to scraper interface


eransakal:
hi @oriStav, Thanks for your contribution. did you see #1064. Isracard and amex are sharing the same base class and from my testing it solved the WAF issue i had in amex.

OriStav:
Hi @eransakal, thanks for flagging #1064 — I hadn't noticed it. I've updated my branch to use that approach as the default  and my previous XLSX download approach is now a fallback.

imrman:
*(Note: I debugged this issue and drafted this comment with the help of my AI pair programmers, Gemini and Codex, while working on a local finance project).*

Hey everyone, thanks for the hard work on these WAF bypasses. 

I've been heavily debugging Amex and Isracard stability in my own project, and the AI uncovered an architectural finding regarding the `amex` and `isracard` base class sharing that might explain some of the erratic "Unknown errors" or post-login failures you are seeing with Amex.

**Our finding: Cross-origin session pollution.**
Initially, our local wrapper used the Isracard portal to "warm up" the browser connection before attempting the native Amex login (since they share infrastructure). However, browserless traces showed that doing this cross-origin pre-navigation actually polluted the shared session with CORS failures and `digital.isracard.co.il/http-errors` *before* the native Amex login even started. 

When we isolated the warmup and navigation to stay **strictly on the Amex origin** (and entirely removed the Isracard portal pre-navigation from the Amex path), our Amex syncs stabilized significantly. 

I'm sharing this in case the base class implementation here is accidentally carrying over Isracard-specific DOM manipulations or portal warmups into the Amex flow. If so, isolating those origins might clear up some of the headless blocks you are currently attributing purely to Cloudflare/WAF. Hope this helps!

PR #1124, opened 2026-06-13
fix(visa-cal): improve change password detection

daniel-hauser:
Copilot on behalf of @daniel-hauser

The Visa Cal website updated how it handles expired passwords. Instead of showing an .err-desc element with 'להחליף סיסמה', it now navigates the login iframe to /change-password and renders a dedicated change-password Angular component.

hasChangePasswordForm() now checks (first match wins):
1. Frame URL contains /change-password
2. change-password Angular component present
3. .change-password-title element present
4. .change-password-subtitle containing 'הגיע הזמן לסיסמה חדשה'
5. Legacy fallback: .err-desc containing 'להחליף סיסמה'

### DIFF
--- a/src/scrapers/visa-cal.ts
+++ b/src/scrapers/visa-cal.ts
@@ -31,6 +31,8 @@ const SSO_AUTHORIZATION_REQUEST_ENDPOINT = 'https://connect.cal-online.co.il/col
 
 const InvalidPasswordMessage = 'שם המשתמש או הסיסמה שהוזנו שגויים';
 const ChangePasswordMessage = 'להחליף סיסמה';
+const ChangePasswordSubtitle = 'הגיע הזמן לסיסמה חדשה';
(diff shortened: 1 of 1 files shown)


daniel-hauser:
Copilot on behalf of @daniel-hauser

Thanks for the review!

**Bug 1 (getLoginFrame throws):** The unguarded `getLoginFrame()` call is a pre-existing pattern — `hasInvalidPasswordError()` has the same issue. That said, it's a fair point and cheap to fix, so I've wrapped the frame-dependent checks in a try/catch that logs and returns false.

**Bug 2 (unscoped frame URL match):** Good catch. Scoped the URL check to frames on `connect.cal-online.co.il` so it won't false-positive on unrelated frames.

Both addressed in the latest force-push.

baruchiro (approved):

PR #1126, opened 2026-06-15
fix(one-zero): bypass Cloudflare bot-detection with Android TLS fingerprint

hadaromer:
## Problem

The One Zero scraper has been failing with HTTP 403 from Cloudflare's bot-management on `identity.tfd-bank.com`. The root cause is that Node.js's built-in `fetch` produces a **TLS JA3 fingerprint** that Cloudflare identifies as automated/non-mobile traffic and blocks — regardless of the HTTP headers sent.

Steps that fail without this fix:
1. `POST /v1/devices/token` → 403 (Cloudflare block page)
2. OTP is never sent → scraper exits with a generic error

## Solution

Add `src/helpers/mobile-fetch.ts` — a thin fetch wrapper using **undici** with the **Android OkHttp 4.x cipher-suite order**. This produces a TLS fingerprint that matches the One Zero mobile app and passes Cloudflare's bot check.

Switch all One Zero identity-server and GraphQL calls to the new `mobileFetchPost` / `mobileFetchGraphql` helpers. The rest of the codebase is unchanged.

**Also fixed:** a pre-existing double-slash bug — `IDENTITY_SERVER_URL` had a trailing slash and every path started with `/`, producing URLs like `/v1//devices/token`.

## Verification

Tested each step of the auth flow against the live API:

| Step | Endpoint | Before | After |
|---|---|---|---|
| 1 | `POST /v1/devices/token` | 403 Cloudflare | ✅ 200 + deviceToken |
| 2 | `POST /v1/otp/prepare` | blocked | ✅ 200 + otpContext |
| 3 | `POST /v1/otp/verify` | blocked | ✅ 200 + otpToken |
| 4 | `POST /v1/getIdToken` | blocked | ✅ 200 + idToken |
| 5 | `POST /v1/sessions/token` | blocked | ✅ 200 + accessToken |

## Changes

- **`src/helpers/mobile-fetch.ts`** *(new)* — undici-based fetch with Android OkHttp TLS profile
- **`src/scrapers/one-zero.ts`** — use `mobileFetchPost`/`mobileFetchGraphql`; fix `IDENTITY_SERVER_URL` double-slash
- **`package.json`** — add `undici` as a runtime dependency

`undici` is the official Node.js HTTP client (maintained by the Node.js core team) and is already the underlying implementation of the global `fetch` in Node 18+. Adding it as an explicit dependency to customise its TLS agent is the minimal-footprint approach.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,8 @@
         "debug": "^4.3.2",
         "moment": "^2.22.2",
         "moment-timezone": "^0.5.37",
-        "puppeteer": "^24.40.0"
+        "puppeteer": "^24.40.0",
+        "undici": "^6.27.0"
       },
       "devDependencies": {
         "@babel/cli": "^7.4.4",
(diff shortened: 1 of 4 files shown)


hadaromer:
Superseded by #1127 which combines this TLS fix with an opt-in credit card feature on top of it.

PR #1128, opened 2026-06-15
feat(one-zero): fix Cloudflare TLS blocking and add opt-in credit card scraping

hadaromer:
## Summary

Two related changes for the One Zero scraper:

**1. Fix Cloudflare bot-detection (TLS fingerprint)**

One Zero's API is behind Cloudflare, which blocks Node.js's native `fetch` based on its TLS JA3 fingerprint. This adds a new helper `src/helpers/mobile-fetch.ts` that uses `undici` with an Android OkHttp 4.x cipher suite order, bypassing the block. All One Zero HTTP calls now go through this helper.

**2. Opt-in credit card transaction scraping**

Adds `oneZero:includeCreditCards` to `OptInFeatures`. When enabled, the scraper calls `cardsDetailsV3` to list all active credit cards in the portfolio, then `cardActivityV3` per card per billing month to collect individual transactions — returned as additional accounts named `*<last4digits> <cardholder name>` (e.g. `*1234 ישראל ישראלי`).

- Installment transactions are mapped with correct `installments.number` / `installments.total` metadata
- Foreign-currency transactions populate `originalAmount`/`originalCurrency` from `debitAmount` (ILS) and `originalAmount` (foreign) fields respectively
- The API uses `BOOKED` (not `COMPLETED`) for settled transactions; mapped correctly to `TransactionStatuses.Completed`

> **Note:** One Zero bank account movements include a monthly lump-sum settlement entry (חיוב ישראכארט) representing the same spending as the individual card transactions. Users who enable this opt-in should be aware of potential duplication.

## Usage

```ts
const scraper = createScraper({
  companyId: CompanyTypes.oneZero,
  startDate,
  optInFeatures: ['oneZero:includeCreditCards'],
});
```

## Changes

- `src/helpers/mobile-fetch.ts` — new helper with Android TLS cipher suite via `undici`
- `src/scrapers/one-zero.ts` — use `mobileFetchPost`/`mobileFetchGraphql`; add `fetchCreditCardAccounts()` method; check opt-in flag in `fetchData()`
- `src/scrapers/one-zero-queries.ts` — add `GET_CARDS_DETAILS` and `GET_CARD_ACTIVITY` queries
- `src/scrapers/interface.ts` — add `'oneZero:includeCreditCards'` to `OptInFeatures`

## Test plan

- [x] `npm run build` passes with no lint or type errors
- [x] One Zero bank account movements scrape successfully (previously returned 403)
- [x] Scraping with `optInFeatures: ['oneZero:includeCreditCards']` returns additional accounts per card
- [x] Settled transactions show `status: completed` (API returns `BOOKED`)
- [x] Installment transactions have `type: 'installments'` and correct `installments` metadata
- [x] Foreign-currency transactions have `originalAmount`/`originalCurrency` populated
- [x] Multiple cardholders on the same portfolio are distinguishable by account name


Isaac-D20:
Regarding the Cloudflare bot-detection issue, it was solved in my case by using the **`axios`** package instead of the native fetch.
Here's my forked version: https://github.com/Isaac-D20/israeli-bank-scrapers/blob/master/src/helpers/fetch.ts

hadaromer:
Thanks for the suggestion! I actually tried axios first too — it does work, but I don't think it's fixing the real problem, just sidestepping it by luck.

Both `fetch` (undici) and `axios` (Node's `http`/`https`) use Node's default TLS stack, and neither sets an explicit cipher suite. Axios getting through is just a side effect of its default agent producing a slightly different cipher order than undici's — not a deliberate fingerprint match. That could break on any Node or axios version bump, or the moment Cloudflare tightens its rule.

`mobile-fetch.ts` instead explicitly sets ciphers to match a real Android OkHttp 4.x client — a deliberate, documented fingerprint rather than an accident of defaults, so it should hold up better and is easy to update if Cloudflare changes things again.

Happy to add axios as a fallback if there's value in having both, but I'd lean toward keeping the explicit TLS approach as the primary fix.

PR #1133, opened 2026-06-30
fix(mizrahi): fetch transactions via direct API call, variant-agnostic (OnlineApp & OnlineAppPilot)

shakeapp1:
## Problem

Two issues break the Mizrahi Tefahot scraper on the current web-banking UI:

1. **Transactions fetch fails headless.** `get428Index` is fired from an in-page iframe **only in response to a real user click**. The pristine code clicks a menu link and `waitForRequest`s for `get428Index`; in headless Chromium that click never triggers the iframe request, so `Promise.any([...])` rejects → `errorType: GENERIC` ("All promises were rejected").

2. **The account is served behind two path variants.** The same account is reachable as the standard **OnlineApp** (API under `/Online/…`) or as **OnlineAppPilot** (API under `/OnlinePilot/…`). Which one a session gets depends on the login entry point, **not on a date** — after login the account is identical, and both variants are live in the field at the same time. Any fix must therefore work on **either** variant.

## Fix

After login the session's `mizrahixsrftoken` is sent automatically on every authenticated `…/api/…` request (`keepAlive`, `Get428ODS`, …), on whichever variant the session is on. This PR captures that token from any such request, **derives the API base from the request that carried it**, and issues `get428Index` directly (deterministic, headless-safe). The matcher keys on the **token header only — never on the `Online`/`OnlinePilot` segment name** — with the two known bases as ordered fallbacks, so it works on either variant and survives a future rename of the segment.

- Login detection accepts `/OnlineApp(Pilot)?/`.
- `get428Index` and `getMaherBerurimSMF` are issued against the derived base (with fallback), so memos populate on both variants.

> ⚠️ Note for reviewers: hardcoding the paths to `/OnlinePilot/` (or to `/Online/`) is **not** a valid fix — it just moves the 90s-timeout / GENERIC failure onto every user on the other variant. Keying on the session token avoids that.

## Verification

- Live **standard** account (`/OnlineApp/` → derived base `…/Online`): full transaction list returned; `getMaherBerurimSMF` returns memos for credits and debits.
- Live **Pilot** account (`/OnlineAppPilot/` → derived base `…/OnlinePilot`): verified live on June 30th using the same direct POST approach.
- Offline unit check of base-derivation + candidate ordering + token predicate (standard, pilot, and a hypothetical renamed segment).

### DIFF
--- a/src/scrapers/mizrahi.ts
+++ b/src/scrapers/mizrahi.ts
@@ -77,7 +77,7 @@ type MoreDetails = {
 const BASE_WEBSITE_URL = 'https://www.mizrahi-tefahot.co.il';
 const LOGIN_URL = `${BASE_WEBSITE_URL}/login/index.html#/auth-page-he`;
 const BASE_APP_URL = 'https://mto.mizrahi-tefahot.co.il';
-const AFTER_LOGIN_BASE_URL = /https:\/\/mto\.mizrahi-tefahot\.co\.il\/OnlineApp\/.*/;
+const AFTER_LOGIN_BASE_URL = /https:\/\/mto\.mizrahi-tefahot\.co\.il\/OnlineApp(Pilot)?\/.*/;
 const OSH_PAGE = '/osh/legacy/legacy-Osh-Main';
 const TRANSACTIONS_PAGE = '/osh/legacy/root-main-osh-p428New';
 const TRANSACTIONS_REQUEST_URLS = [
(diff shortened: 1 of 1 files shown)


shakeapp1:
Updated this PR to be variant-agnostic. The earlier revision hardcoded `/OnlinePilot/` for the token wait, `get428Index`, and the more-details endpoint — that would break every user still on the standard `/OnlineApp/` (`/Online/api/`) door with a 90s `waitForRequest` timeout (`errorType: GENERIC`).

This revision instead grabs the session `mizrahixsrftoken` from any authenticated `…/api/…` request and derives the API base from that request, with the two known bases as ordered fallbacks. It works on both the standard OnlineApp and the OnlineAppPilot variants (both are live in the field), and survives a future rename of the path segment. Verified live against both a standard and a Pilot account.

PR #1026, opened 2025-12-23
feat(hapoalim): add forex, savings, & investment accounts

Skeli789:
- Adds support for scraping transaction history from Bank Hapoalim foreign currency accounts
- Adds support for scraping Bank Hapoalim savings account balances
- Adds support for scraping Bank Hapoalim investment account holdings

Tested locally. New accounts in output look like:
```json
    {
      "accountNumber": "12-XXX-XXXXXX-USD",
      "balance": 1000.00,
      "currency": "USD",
      "txns": [
        {
          "type": "normal",
          "identifier": 10017,
          "date": "2025-10-02T21:00:00.000Z",
          "processedDate": "2025-10-15T21:00:00.000Z",
          "originalAmount": 1000.00,
          "originalCurrency": "USD",
          "chargedAmount": 1000.00,
          "description": "העברה מחו\"ל",
          "status": "completed",
          "memo": "Bob Smith"
        }
      ]
    },
    {
      "accountNumber": "12-XXX-XXXXXX-3382501",
      "savingsAccount": true,
      "balance": 10000.00,
      "txns": []
    },
    {
      "accountNumber": "12-XXX-XXXXXX-investment",
      "balance": 149.98,
      "currency": "ILS",
      "savingsAccount": true,
      "txns": [],
      "securities": [
        {
          "name": "POALIM",
          "symbol": "POLI",
          "volume": 2,
          "value": 149.98,
          "currency": "ILS",
          "changePercentage": -1.04,
          "profitLoss": -1.58
        }
      ]
    }
```

PR #1149, opened 2026-07-26
feat(leumi): add foreign currency accounts

geriamir:
Leumi exposes foreign currency (forex) accounts on a separate page that the scraper never visited, so balances and transactions held in USD/EUR/etc. were invisible to consumers.

This follows the approach @Skeli789 took for Hapoalim in #1026: the forex accounts are appended to the `accounts` array returned by the regular `scrape()` rather than exposed through a new API, so existing consumers pick them up with no code changes. No changes to `transactions.ts` were needed, since the optional `currency` field it already carries is enough.

Leumi already had savings account support, so this mirrors that structure (`fetchForeignCurrencyAccounts` alongside `fetchSavingsAccounts`, hooked into `fetchData`).

### Details

Each account is keyed as `<accountNumber>-<currency>` to keep it distinct from the shekel account it hangs off.

A few things that are specific to this page and worth calling out for review:

- **Accounts are fetched sequentially.** They all share a single page and the flow is select account -> select period -> submit, so fetching them concurrently races on the shared form.
- **The selection is submitted before the currency and balance labels are read.** Those labels describe whichever account is currently rendered, so reading them first yields the previous account's values.
- **The currency arrives as a Hebrew label, not an ISO code.** It is matched against a lookup table after stripping quotes, because the punctuation is not stable. Longest label wins, otherwise `דולר קנדי` would match the `דולר` entry. An unrecognised label skips the account rather than defaulting it to shekels.
- **Balance comes from the running balance of the most recent row**, falling back to the opening balance when the range is empty. The table's sort direction is detected by comparing the first and last rows rather than assumed.
- **The whole fetch is wrapped** so a forex failure can never fail the main scrape, and empty accounts are skipped.

### Testing

`npm run lint`, `tsc --noEmit` and the full jest suite pass (18 suites).

Verified end to end against a live Leumi account with a USD forex account, over a 3 month window. The scrape returned the regular shekel account unchanged (74 transactions) plus the forex account, with no invalid dates and no NaN amounts:

```
account       currency  balance  txns
<redacted>    USD       1990     2

  18/06/2026  העברה במט"ח    +29990 USD
  22/06/2026  העברת כספים    -28000 USD
```

29990 - 28000 = 1990, so the running balance taken from the latest row agrees with the transactions, which is the balance question I had flagged. Date parsing (`DD/MM/YY`), the credit/debit sign convention and the `identifier` reference column all check out against real data.

One finding worth highlighting: the live page returns the dollar label as `דולר ארה"ב`, with a real double quote. An exact string comparison against the apostrophe form silently falls through to shekels, which is why the matching here normalises quotes first.

The one thing I still cannot verify is the **non-USD labels** — the test account only holds USD, so the EUR/GBP/etc. entries in the lookup table are inferred rather than observed. An unrecognised label is skipped with a debug message, so it fails safe rather than mislabelling, but if anyone has a EUR or GBP Leumi forex account I would appreciate confirmation of the exact label text.

Happy to adjust to match whatever shape you settle on in #1026.

### DIFF
--- a/.claude/skills/pr-safety-review/SKILL.md
+++ b/.claude/skills/pr-safety-review/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: pr-safety-review
+description: Give a fast, reviewer-facing verdict on whether a PR or diff in this repo is safe to merge — SAFE, RISKY, or BREAKING — instead of a long research writeup. Checks for (1) unnecessary/stale comments, (2) whether existing working code paths still behave identically (backward compatibility), and (3) whether any new feature added alongside the core scraping logic degrades gracefully instead of risking the whole scrape on failure. Use this whenever the user asks to review a PR, look at a diff, check a pull request for breaking changes, asks "is this safe to merge/approve", pastes a GitHub PR URL/number and asks what to watch out for, or says they don't have time to dig into a PR themselves and just want the important part. Trigger even if they only give a PR number or link with no other instructions — that alone means they want this fast verdict, not an essay.
+---
+
+# PR Safety Review
+
+A maintainer of this repo reviews a constant stream of community PRs (mostly scraper fixes/features) and does not have time to re-derive the control flow of every changed function by hand. This skill exists to do that derivation once, carefully, and hand back only the part a reviewer actually needs to act on: **does this change anything that already worked, and if it adds something new, can that new thing fail without taking the rest of the scraper down with it.**
+
+The output is a verdict, not a report. If you catch yourself writing multiple paragraphs of narration about what you read, stop — compress it into the findings list. The reviewer will ask follow-up questions if they want the detail behind a specific finding.
+
+## Step 1 — Get the actual diff
+
+Don't review from the PR title/description alone — authors describe intent, not always the resulting control flow. Fetch:
+- The diff itself (`get_diff` / `git diff`).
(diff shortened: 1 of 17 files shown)


Skeli789:
@baruchiro can you please review?

baruchiro:
It is a lot of code with a long description. I should find a better time to review it.

geriamir:
@baruchiro we are starting to add more features to each scraper - should we consider a way to distribute these features across more files? so it will be more readiable instead of huge scraper files?
I have another feature pending for Leumi (investment accounts) so it will be bigger ;)

geriamir:
@baruchiro  when do you think you can go over this PR?

geriamir:
kind ping @baruchiro :)

baruchiro:
סליחה ממש, אני לא מצליח להתקרב לפה אפילו, ממש מתנצל.

baruchiro (commented):
Here is my suggestion, after I saw this and other PRs:
- Moving the leumi-related files into subfolder `src/scrapers/leumi/`
- add a file like `leumi-forex` to extract all the helper functions as `parseForeignAmount` into it.
- Write a test file on it, with real examples, which will give as the infra for TDD, once we see another example in Leumi, we will be able to add it as test.

geriamir:
@baruchiro done. I also tested it with my account and it works

baruchiro (approved):

PR #1154, opened 2026-08-12
feat(isracard): add credit card balances

Skeli789:
Continuation from https://github.com/eshaham/israeli-bank-scrapers/pull/1138 and https://github.com/eshaham/israeli-bank-scrapers/pull/1141. This PR adds credit card balances and frames to Isracard and Amex.

Local Isracard test results:
```
{
  "success": true,
  "accounts": [
    {
      "accountNumber": "8539",
      "txns": [],
      "balance": 0,
      "cardFrame": 0
    },
    {
      "accountNumber": "8742",
      "txns": [],
      "balance": -10847.5,
      "balanceDate": "2026-09-10T00:00:00",
      "cardFrame": 23500
    }
  ]
}
```

Local Amex test results:
```
{
  "success": true,
  "accounts": [
    {
      "accountNumber": "7392",
      "txns": [],
      "balance": -9564.99,
      "balanceDate": "2026-08-10T00:00:00",
      "cardFrame": 15500
    }
  ]
}
```

PR #1159, opened 2026-08-26
feat(isracard,amex): migrate to Isracard Group's DigitalV3 API

geriamir:
## Summary

Isracard's old browser-driven scraping flow against `digital.isracard.co.il` started hitting bot detection. Isracard and Amex Israel turn out to run on the same Isracard Group backend, so this migrates both scrapers to the group's newer JSON API for card discovery and transaction fetching, while keeping the existing login flow unchanged.

* New shared base `src/scrapers/base-isracard-group.ts`:
  - Login stays on `digital.isracard.co.il` / `he.americanexpress.co.il` (`ValidateIdData` + `performLogonI`), unchanged from the old flow.
  - `GetCardList` on `web.isracard.co.il` / `web.americanexpress.co.il` (`/ocp/transactions/DigitalV3.Transactions/GetCardList`) discovers all cards on the account, filtered by company code — this replaces the old single-card-per-credential model with proper multi-card support.
  - `GetMonthlyBilling` fetches each card's real effective billing date per month, used for accurate `processedDate`/`balanceDate` (previously approximated as the 1st of the month).
  - `GetTransactionsList` merges pending approvals, settled vouchers, and out-of-statement immediate-debit vouchers into one transaction list per card/month.
* `isracard.ts` and `amex.ts` are now thin subclasses of the shared base, differing only in base URLs and company code (mirrors the pattern the old `base-isracard-amex.ts` used).
* New data now populated that the old flow didn't expose: `category` (from `branchCodeDescription` for pending transactions, `transactionDescription` for settled ones), and account `balance`/`cardFrame` from the card's credit limit data.
* Amex's login was separately being blocked by a Cloudflare rule that checks `navigator.webdriver` (Chromium's own automation flag) — reproduced identically against the old, unmodified `base-isracard-amex.ts` code, confirming it predates this change. Fixed by patching that flag via `page.evaluateOnNewDocument` before navigating to the login page. This also removes one potential future failure point for Isracard, even though Isracard wasn't currently blocked by it.
* Removed the now-unused `base-isracard-amex.ts`.

## Test plan

Verified end-to-end against real Isracard and Amex accounts (login → card discovery → effective billing dates → transaction fetching, across multiple months and multiple cards per account). Also ran:
- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm test` (all other scraper suites unaffected; isracard/amex live-API tests pass with real credentials)


Skeli789:
Make sure this is updated with the credit card balance info from https://github.com/eshaham/israeli-bank-scrapers/pull/1154

geriamir:
Good catch on the timing (this branch predates #1154) - but the balance/frame data is already covered here, just sourced differently: `getCardBalance`/`getCardFrame`/`getCardBalanceDate` in `base-isracard-group.ts` (added in this PR) read it straight from the DigitalV3 `GetCardList` response's `limitData.limitUsed`/`creditLimitAmount` and `cardChargeNext.billingDate` fields, rather than #1154's regex-over-Hebrew-page-text approach - both end up populating the same `balance`/`balanceDate`/`cardFrame` fields on the account.

What was actually missing was test coverage: this PR had deleted `base-isracard-amex.test.ts` (which covered #1154's balance parsing) along with the file it tested, without adding anything in its place. Just pushed unit tests for the new base's balance derivation and transaction mapping to close that gap.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

geriamir:
@baruchiro  do you think we can get this in?

src/scrapers/base-isracard/base-isracard-utils.test.ts
```diff
@@ -0,0 +1,199 @@
+import moment from 'moment';
```
baruchiro, line 1: like I suggested to other one, if you can split the scraper from its helpers, so in final you will have a folder for base-isracard, and inside you will have the scraper, the `base-isracard-utils` and `base-isracard-utils.test` it will be easier to review.

I think it was already done for Leumi

geriamir: Done — split into `src/scrapers/base-isracard/`: `base-isracard.ts` for the scraper class and Puppeteer-driven fetch calls, `base-isracard-utils.ts` for the pure balance/frame/date derivation and transaction-conversion helpers, with `base-isracard-utils.test.ts` alongside them. Also fixed `component_owners.yml`, which still pointed at the old (already-deleted) `base-isracard-amex.ts` path.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

OriEylon:
Tested this branch against a real Isracard account (`--months 1`, 3 cards). Login, card discovery, and transaction fetch all worked cleanly — no 429s.

However, the per-month transaction list has widespread duplicates. `GetTransactionsList`'s merge of pending approvals + settled vouchers + immediate-debit vouchers doesn't dedupe charges that appear in more than one of those lists: a charge that has both a pending-approval entry and a later settled-voucher entry for the same amount/date comes back as **two separate transaction rows** with different generated ids, rather than being collapsed into one. Shape of it (values anonymized):

```
<merchant>, <amount> ILS, accountNumber <card-last-4>, processedDate <date>
  id <hash-a>... category: null              (from pending approvals)
  id <hash-b>... category: "<hebrew category>"  (from settled vouchers)
```

On the 1-month/43-transaction test pull, 31 of 43 transactions (72%) were affected — i.e. most settled transactions doubled. A full-year scrape on this branch would roughly double transaction counts/totals for anyone consuming the output.

The two rows for the same underlying charge are distinguishable in this sample by `category` being null vs. populated (pending vs. settled), so a merge/dedupe step keyed on (account, amount, processedDate) — preferring the settled/categorized entry — looks like it would fix it, but wanted to flag before this gets merged since the duplication isn't obvious from a quick smoke test (row counts still look plausible, "amounts double" only shows up when reconciling against a real statement).

Happy to share more sample data privately if useful.

geriamir:
Good catch, thanks for testing against a real account. Confirmed: `GetTransactionsList`'s response can list the same charge in both `approvals.approvedTransactions` (pending) and the voucher lists (settled) at the same time, so it was showing up twice.

Fixed in `4102142`: added `isApprovedTransactionSettled` (matches on purchase date + amount + currency + business name) and filter out any approval that already has a matching settled voucher before converting, keeping the settled/voucher entry (which has the richer data — category, installments, etc.). Added unit tests covering the match/no-match cases in `base-isracard-utils.test.ts`.

Happy to look at your sample data if you still see duplicates after this — the match key doesn't include time-of-day since vouchers don't always carry it, so let me know if that turns out to be too loose in practice.

geriamir:
@OriEylon Can you test the recent changes?

PR #1153, opened 2026-08-12
fix(visa-cal): more credit card balance issues

Skeli789:
Fixes Cal credit card balances and frames not loading for some people because of how the data in their account is structured.

Local test results:
```
{
  "success": true,
  "accounts": [
    {
      "txns": [],
      "balance": -327,
      "balanceDate": "2026-09-02T00:00:00",
      "accountNumber": "9106",
      "cardType": "bankIssued",
      "cardFrame": 7500
    }
  ]
}
```

Debug logs
```
  israeli-bank-scrapers:visa-cal   result: {
  israeli-bank-scrapers:visa-cal     calIssuedCards: null,
  israeli-bank-scrapers:visa-cal     bankIssuedCards: {
  israeli-bank-scrapers:visa-cal       fictiveMaxAccAmt: 7075,
  israeli-bank-scrapers:visa-cal       frameLimitForCardAmount: 7500,
  israeli-bank-scrapers:visa-cal       totalUsageAmountForAccountManagementLevel: 327,
  israeli-bank-scrapers:visa-cal       usageAmountForExternalCards: 98,
  israeli-bank-scrapers:visa-cal       nextTotalDebitDateForAccount: '2026-09-02T00:00:00',
  israeli-bank-scrapers:visa-cal       nextTotalDebitForAccount: 327,
  israeli-bank-scrapers:visa-cal       currencyCode: 3,
  israeli-bank-scrapers:visa-cal       currencySymbol: '₪',
  israeli-bank-scrapers:visa-cal       summary: [Object],
  israeli-bank-scrapers:visa-cal       accountComments: [Array],
  israeli-bank-scrapers:visa-cal       accountLevelFrames: [],
  israeli-bank-scrapers:visa-cal       cardLevelFrames: [],
  israeli-bank-scrapers:visa-cal       accountExceptionalCards: [Array]
  israeli-bank-scrapers:visa-cal     }
  israeli-bank-scrapers:visa-cal   },
  israeli-bank-scrapers:visa-cal   statusDescription: null,
  israeli-bank-scrapers:visa-cal   statusTitle: null,
  israeli-bank-scrapers:visa-cal   statusCode: 1,
  israeli-bank-scrapers:visa-cal } +72ms
```

PR #1162, opened 2026-09-06
feat: add pr-safety-review skill for fast PR safety verdicts

baruchiro:
## Summary

Adds a Claude Code skill, `pr-safety-review`, for reviewing PRs/diffs in this repo and giving a fast, reviewer-facing verdict instead of a long writeup.

For a given PR or diff, it runs three checks and reports a single `SAFE | RISKY | BREAKING` verdict with tagged, file:line-anchored findings:

- **comments** — flags unnecessary comments (restates the code, references an issue/PR instead of a durable "why", stale relative to the diff)
- **compat** — walks the control flow to classify the change as safe (only touches previously-broken/unreached paths), risky (an already-working path's behavior changes for reachable real inputs), or breaking (a public type/return shape/signature changes in a way existing consumers would notice)
- **feature-safety** — for new capabilities bolted onto an existing scraper flow (an extra field, an extra network fetch), checks that a failure in the new part degrades gracefully (contained by try/catch at the right scope, fails to `undefined` rather than a fabricated value, doesn't mutate shared state the core flow still needs) instead of risking the whole scrape

It also checks CI status/open review threads for real PRs, and flags when a changed file is a shared base class used by other scrapers (e.g. `base-isracard-amex.ts`, `base-beinleumi-group.ts`) so the blast radius of a compat finding is accurate.

## How it was built and validated

Drafted against two real PRs from this repo (#1153, #1154) and a synthetic diff seeded with a breaking rename, an unguarded new-feature `await`, and unnecessary comments, then benchmarked with-skill vs. a no-skill baseline across two iterations:

- Iteration 1: 83% pass rate with-skill vs. 38% baseline
- Iteration 2 (after fixing gaps iteration 1 surfaced): 93% pass rate with-skill

Eval definitions, fixtures, and expected outcomes are included under `evals/` for anyone who wants to re-run or extend them.

## Test plan

- [x] Ran against real PRs #1153 and #1154 in this repo and against a synthetic seeded-bug diff; verified outputs against a fixed set of assertions per case
- [x] Confirmed the "shared base class" examples cited in the skill against the actual codebase (`grep -rl` for importers) rather than assuming
- [ ] No automated CI for the skill itself — it's a Markdown instruction file, not application code, so there's nothing for this repo's existing test suite to run against

https://claude.ai/code/session_01QYsa6xivQwaSq3cdiqVDdp

---
_Generated by [Claude Code](https://claude.ai/code/session_01QYsa6xivQwaSq3cdiqVDdp)_


.claude/skills/pr-safety-review/evals/evals.json
baruchiro, line 1: is this should be committed?

baruchiro: Yes, intentionally — `evals/evals.json` (plus `evals/fixtures/bad-diff.txt`) is the skill's own test suite: 3 prompts with expected-output assertions, used to benchmark this skill against a no-skill baseline before merging (results in the PR description). Committing it means anyone who later edits the skill can re-run the same cases to check they didn't regress, rather than having to reconstruct the eval setup from scratch — similar to committing test fixtures alongside application code.

The actual eval *run output* (transcripts, grading, the benchmark scorecard) lives in a separate `pr-safety-review-workspace/` directory that's gitignored (see the `.gitignore` diff in this PR) — only the reusable test definitions are checked in, not the scratch artifacts from running them.

Happy to drop it if you'd rather keep the skill directory to just `SKILL.md`, let me know.

---
_Generated by [Claude Code](https://claude.ai/code)_

.gitignore
```diff
@@ -64,3 +64,6 @@ lib/
 .idea
 .vscode
 .DS_Store
+
+# skill-creator eval workspaces (scratch, not part of the skill itself)
+.claude/skills/*-workspace/
```
baruchiro: remove

.claude/skills/pr-safety-review/evals/fixtures/bad-diff.txt
baruchiro, line 1: is this referenced from the skill?

baruchiro: Not from `SKILL.md` itself — it's referenced from `evals/evals.json` (eval id 3, `synthetic-breaking-and-unsafe-feature`), both in the prompt text and its `files` field. It's the fixture diff that eval feeds to the skill to test the "does it actually catch a breaking change + an unsafe new feature" case, since I didn't have a real PR handy with those exact problems.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #972, opened 2025-08-11
feat(leumi): add investments

geriamir:
# Add Investment Portfolio Support for Leumi Bank 🏦📈

## Overview
This PR adds comprehensive investment portfolio scraping capabilities to the Leumi bank scraper, enabling users to retrieve their investment holdings and transaction history alongside their regular banking transactions.

### 🆕 New Files
- **`src/investments.ts`** - Core investment data structures and interfaces

### 🔧 Modified Files
- **`src/scrapers/leumi.ts`** - Enhanced with investment portfolio scraping functionality
- **`src/scrapers/interface.ts`** - Updated to support portfolio data in scraper results

### 📊 Investment Data Structures
Added three new interfaces to support comprehensive investment tracking:

#### `Portfolio`
- `portfolioId`: Unique identifier for the portfolio
- `portfolioName`: Human-readable portfolio name
- `investments[]`: Array of current investment holdings
- `transactions[]`: Array of investment transaction history

#### `Investment`
- `paperId`: Unique paper/security identifier
- `paperName`: Security name
- `symbol`: Trading symbol
- `amount`: Number of shares/units held
- `value`: Current market value
- `currency`: Currency denomination

#### `InvestmentTransaction`
- Complete transaction details including execution date, price, tax information
- Links to specific securities via `paperId` and `symbol`
- Supports multi-currency transactions

### 🔍 Leumi Scraper Enhancements
The Leumi scraper now includes:

#### New Trading URLs
- `LEUMI_TRADING_URL`: Portfolio holdings page
- `LEUMI_TRADING_HISTORY_URL`: Transaction history page

#### Core Investment Methods

**`fetchPortfolios(startDate)`**
- Navigates to Leumi's trading platform
- Intercepts XHR/fetch requests to extract portfolio data
- Handles multiple response types (config, statement, history)
- Returns complete portfolio with investments and transactions

**`fetchPortfolioTransactions(startDate)`**
- Specifically handles transaction history retrieval
- Implements custom date picker navigation for historical data
- Supports configurable start date filtering
- Returns array of `InvestmentTransaction` objects

**`extractPortfolios(response, portfolios[])`**
- Parses portfolio metadata from API responses
- Extracts portfolio IDs and names

**`extractPortfolioInvestments(response, investments[])`**
- Parses current holdings from trading platform API
- Handles currency conversion (currently defaults to ILS)
- Extracts paper details, amounts, and current values

**`extractPortfolioTransactionsFromResponse(response)`**
- Processes historical transaction data
[...]

PR #1166, opened 2026-09-10
feat(leumi): add investment portfolio accounts

geriamir:
Adds Leumi trading-portfolio holdings and order history as investment accounts, following #1149 into the trading side of the account: same "no new top-level shape" approach, appending to the existing `accounts` array.

### Schema

Aligned with #1026's convention rather than my own earlier, closed #972 (bespoke `Portfolio`/`Investment`/`InvestmentTransaction` types + a new top-level `portfolios` field, which never got real review): an investment portfolio is a regular `TransactionsAccount` entry (`accountNumber` suffixed `-investment`, `savingsAccount: true`) with a new `securities: Security[]` field, matching the `Security` interface #1026 already introduced (`name?`, `symbol`, `volume`, `value`, `currency?`, `changePercentage?`, `profitLoss?`).

### Implementation

New `src/scrapers/leumi/leumi-investments.ts`. Rather than driving the trading page's UI (an Angular Material date picker, in particular), it calls the same three endpoints the page itself calls, directly:

* `GET /lti/lti-app/api/config` - lists portfolios
* `GET /lti/lti-app/api/Trade/Statement?PortfolioIndex=...` - current holdings + `PortfolioValue` (used as the account balance, since it also covers uninvested cash, rather than summing holdings)
* `GET /lti/lti-app/api/Trade/GetOrdersHistory?portfolioIndex=...&FromDate=...&Todate=...` - order history, mapped to `Transaction[]`

All three are plain authenticated GETs (same pattern as the existing savings-account fetch, via `fetchGetWithinPage`), so no browser interaction beyond being logged in is needed. Portfolios are addressed by their index in the `config` response, so multiple portfolios are all fetched (unlike an earlier draft of this that only handled the first one).

One thing worth calling out: order rows carry `ExecutableTotal` signed by *quantity* direction (positive on a buy, negative on a sell) rather than cash-flow direction, so the transaction amount is instead derived from the `TypeOfOperation` field (1 = buy, 2 = sell) - a buy is a negative amount (money leaving the account), a sell positive.

### Testing

`npm run lint`, `tsc --noEmit` and the full jest suite pass. Unit tests cover the pure parsing functions against real (redacted account numbers) response shapes captured from a live account: an empty portfolio, one holding a single money-market fund, and an order-history pair (a fund switch: sell one fund, buy another).

Verified end to end against a live Leumi account via the existing `leumi.test.ts` live-API test (extended here to validate investment accounts the same way it already validates savings/forex accounts).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01W1W6K5gQtDmeQqLUfJQxEq

### DIFF
--- a/src/scrapers/leumi/leumi-investments.test.ts
+++ b/src/scrapers/leumi/leumi-investments.test.ts
@@ -1,0 +1,91 @@
+import { type ScraperOptions } from '../interface';
+import { parseHoldingsResponse, parseOrderHistoryResponse, parsePortfoliosResponse } from './leumi-investments';
+
+describe('parsePortfoliosResponse', () => {
+  test('extracts portfolios from the config response', () => {
+    const data = {
+      data: {
+        user: {
+          Portfolios: [{ PortfolioId: '12345', PortfolioName: 'תיק מסחר' }],
+        },
+      },
+    };
+
+    expect(parsePortfoliosResponse(data)).toEqual([{ PortfolioId: '12345', PortfolioName: 'תיק מסחר' }]);
+  });
+
+  test('returns an empty array when the response has no portfolios', () => {
+    expect(parsePortfoliosResponse({ data: { user: {} } })).toEqual([]);
+    expect(parsePortfoliosResponse({})).toEqual([]);
+    expect(parsePortfoliosResponse(null)).toEqual([]);
+  });
+});
+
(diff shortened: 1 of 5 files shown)


geriamir:
@baruchiro can you go over this PR? its in the same lines of what we've done with Leumi's forex.

PR #1172, opened 2026-09-12
fix(one-zero): present the client certificate Cloudflare mTLS requires

### LINKED ISSUE #1144: OneZero authentication fails with Cloudflare 403 when using Node’s built-in fetch/Undici
## Description

The OneZero scraper fails during `triggerTwoFactorAuth()` because requests to OneZero’s identity API are blocked when sent through Node’s built-in `fetch()` implementation.

The first failing request is:

```http
POST https://identity.tfd-bank.com/v1/devices/token
Content-Type: application/json

{"extClientId":"mobile","os":"Android"}
```

The endpoint returns HTTP 403 with a Cloudflare HTML response:

```text
status: 403 Forbidden
content-type: text/html; charset=UTF-8
server: cloudflare
```

Because `fetchPost()` calls `response.json()` without first checking the response status or Content-Type, the error exposed to the caller is:

```text
SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

This hides the underlying Cloudflare rejection.

## Environment

- `israeli-bank-scrapers`: 6.8.0
- Node.js: 24.18.0
- Bundled Undici: 7.28.0
- Platform: macOS
- OneZero scraper
- Endpoint: `identity.tfd-bank.com`

## Steps to reproduce

A minimal reproduction using Node’s built-in `fetch()`:

```js
const response = await fetch(
  'https://identity.tfd-bank.com/v1/devices/token',
  {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
    },
    body: JSON.stringify({
      extClientId: 'mobile',
      os: 'Android',
    }),
  },
);

console.log(response.status);
console.log(response.headers.get('content-type'));
console.log((await response.text()).slice(0, 100));
```

Observed result:

```text
403
text/html; charset=UTF-8
<!DOCTYPE html>
```

The same request succeeds with curl:

```bash
curl -X POST \
  'https://identity.tfd-bank.com/v1/devices/token' \
  -H 'Content-Type: application/json' \
  -d '{"extClientId":"mobile","os":"Android"}'
```

Observed result:

```text
HTTP 200
Content-Type: application/json
```

The response contains a valid `resultData.deviceToken`.

## Differential diagnosis

The requests below were made from the same machine and public IP:

| Client/request variant | Result |
|---|---:|
| curl using negotiated HTTP/2 | 200 JSON |
| curl forced to HTTP/1.1 | 200 JSON |
| curl with `User-Agent: undici` | 200 JSON |
| Node built-in `fetch()` | 403 Cloudflare HTML |
| Node `fetch()` with curl’s User-Agent | 403 Cloudflare HTML |
| Node `https.request()` | 200 JSON |
| `https.request()` with Undici-like headers | 200 JSON |

This indicates that the failure is not caused by:

- The JSON payload
- HTTP/1.1 versus HTTP/2
- The `User-Agent` header alone
- One of the other visible default headers alone

It appears Cloudflare is distinguishing Undici through lower-level request serialization, connection behavior, or TLS/client fingerprinting.

## Expected behavior

`triggerTwoFactorAuth()` should obtain a OneZero device token and continue to `/otp/prepare`.

If the remote service returns an HTTP or non-JSON error, the scraper should report the status and response type instead of throwing a JSON parsing error.

## Actual behavior

The request receives HTTP 403 HTML, after which `fetchPost()` attempts to parse the HTML as JSON and throws:

```text
Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

## Suggested fix

Use a `node:https`-based JSON request helper for calls to OneZero’s identity service:

```text
https://identity.tfd-bank.com/v1/...
```

This should cover at least:

- `/devices/token`
- `/otp/prepare`
- `/otp/verify`
- `/getIdToken`
- `/sessions/token`

The transport change should preferably be scoped to the OneZero identity client. Other bank scrapers should continue using their existing transport unless they reproduce the same problem.

Additionally, `fetchPost()` should:

1. Read the response as text.
2. Check whether the status is successful.
3. Check whether the Content-Type is JSON.
4. Only then parse the response with `JSON.parse()`.

For example, a rejected response should produce an error similar to:

```text
POST https://identity.tfd-bank.com/v1/devices/token returned
403 Forbidden; content-type=text/html; charset=UTF-8
```

instead of a JSON syntax error.

MatanYadaev:
Fixes #1144

## Problem

Since August 2026 OneZero's API hosts (`identity.tfd-bank.com`, `mobile.tfd-bank.com`) sit behind Cloudflare mTLS. A request without a client certificate gets a 403 HTML page from Cloudflare before it reaches the bank, so the scraper dies on the first call (`/devices/token`) with `Unexpected token '<'`.

It is not a TLS fingerprint issue. curl fails the same way without a certificate and succeeds with it, on the same machine:

```
curl -X POST https://identity.tfd-bank.com/v1/devices/token -H 'Content-Type: application/json' -d '{"extClientId":"mobile","os":"Android"}'
→ 403 text/html

curl --cert mtls_cert.pem --key mtls_key.key ... (same request)
→ 200 application/json
```

## Fix

- `one-zero-client-cert.ts`: the certificate and key the OneZero Android app ships in its APK (`res/raw/mtls_cert.pem`, `res/raw/mtls_key.key`). They are identical in every install and are presented before login, so they identify the app, not a user. Expires 2027-08-05.
- `fetch.ts`: `fetchPost` / `fetchGraphql` take an optional `clientCert`. Node's `fetch` has no client-certificate option, so those requests go through `https.request`. A non-JSON body now throws an error with the status code instead of a JSON parse error (the other ask in #1144).
- `one-zero.ts`: every request passes the certificate.

Same approach as [finance-analysis](https://github.com/tomerroditi/finance-analysis/blob/main/scraper/providers/banks/onezero.py), which has been running on it since August.

## Tested

- `npm run build`, lint and type-check pass. Unit tests for `fetchPost` in `fetch.test.ts` cover the JSON, non-JSON and client-certificate paths.
- Full OneZero flow against the real API: SMS OTP, `getLongTermTwoFactorToken`, login, account list, movements.
- Without the certificate the new error is `sending a request to the institute server returned with status code 403 and a non-JSON body`.


shustinm:
@MatanYadaev thanks for the fix!
it worked for me! 👍

